PR #30, opened 2019-09-10
QA-12029 : added a new filter to check the licence and user the licen…

Yousria:
…ceChecker service to verify the validity


src/main/java/org/jahia/modules/contentmanager/license/LicenseCheckFilter.java
```diff
@@ -0,0 +1,54 @@
...
+ * Filter inserted in the render chain to validate the licence for CMM
+ * @author yousria
+ */
+@Component(service = RenderFilter.class, immediate = true, name = "LicenseCheckFilterCMM")
+public class LicenseCheckFilter extends AbstractFilter {
+
+    public LicenseCheckFilter() {
+        setPriority(1);
+        setApplyOnTemplates("content-manager");
+    }
```
tdraier: Rather use spring file to declare bean and set properties, as spring is already used in this module

PR #41, opened 2019-10-18
BACKLOG-11208 : zip Action in CMM

Yousria:


src/javascript/ContentManager/ContentManager.utils.js
```diff
@@ -59,6 +59,36 @@ function ellipsizeText(text, maxLength) {
...
 
+function getZipName(name) {
+    if (name.lastIndexOf('.') > 0) {
+        return name.substr(0, name.lastIndexOf('.')) + '.zip';
+    }
+
+    return name + '.zip';
+}
+
+function getNameWithoutExtension(name) {
```
cedricmoitrier: I would rename this to "removeFileExtension(filename)"

src/javascript/ContentManager/ContentManager.utils.js
```diff
@@ -59,6 +59,36 @@ function ellipsizeText(text, maxLength) {
...
+    nodes.forEach(node => {
+        let counter = node.name.match(/[0-9]+.zip/g);
+        if (counter !== null) {
+            counter = counter[0].match(/[0-9]+/g);
+            if (counter !== null && counter[0] > max) {
+                max = counter;
+            }
+        }
+    });
+    return parseFloat(max) + 1;
```
cedricmoitrier: Why parseFloat?

src/javascript/ContentManager/ContentManager.utils.js
```diff
@@ -59,6 +59,36 @@ function ellipsizeText(text, maxLength) {
...
+        return name.substr(0, name.lastIndexOf('.'));
+    }
+
+    return name;
+}
+
+function getNewCounter(nodes) {
+    let max = 0;
+    nodes.forEach(node => {
+        let counter = node.name.match(/[0-9]+.zip/g);
```
cedricmoitrier: You could probably make this method more generic, by replacing this line by some code removing the file extension (whatever it is) from the name (ie. calling the method above).

src/javascript/ContentManager/actions/zipUnzip/zipAction.jsx
```diff
@@ -0,0 +1,47 @@
...
+import zipUnzipMutation from './zipUnzip.gql-mutations';
+import {getZipName, getNewCounter, getNameWithoutExtension} from '../../ContentManager.utils';
+
+export default composeActions(requirementsAction, withNotificationContextAction, {
+    init: context => {
+        context.initRequirements({});
+    },
+    onClick: context => {
+        let name = context.node ? context.node.name : context.nodes[0].parent.name;
+        let nameWitoutExtension = getNameWithoutExtension(name);
```
cedricmoitrier: typo: nameWit[h]outExtension

src/javascript/ContentManager/ContentManager.utils.js
```diff
@@ -59,6 +59,36 @@ function ellipsizeText(text, maxLength) {
     return ellipsize(text, maxLength || 100, {chars: [' ', '&']});
 }
 
+function getZipName(name) {
```
cedricmoitrier: I would get rid of the function. It does not bring much value and the naming is quite confusing.
Have the caller use the method below to remove the extension if any, and then append '.zip'

src/main/resources/javascript/locales/en.json
```diff
@@ -77,6 +77,9 @@
         "dropMessage": "Drag and drop a file to import",
         "action": "Import content"
       },
+      "zipUnzip": {
+        "zip": "zip"
```
cedricmoitrier: Value should start with an upper-cased character (ie. should be Zip)

src/javascript/ContentManager/ContentManager.utils.js
Yousria: I used it only to force addition rather than concatenation

PR #43, opened 2019-10-23
BACKLOG-11210 : test for utils functions used by zipUnzip actions

Yousria:
## JIRA

https://jira.jahia.org/browse/BACKLOG-11210

## Description
add tests for utils functions

## Tests

no need to test, CMM needs only to build
## Checklist

- [x] Performance
- [x] Migration
- [x] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/ContentManager/ContentManager.utils.test.js
+++ b/src/javascript/ContentManager/ContentManager.utils.test.js
@@ -1,0 +1,27 @@
+import {removeFileExtension, getNewCounter} from './ContentManager.utils';
+
+describe('zip utils', () => {
+    it('should remove file extension', () => {
+        expect(removeFileExtension('file.jpg')).toEqual('file');
+        expect(removeFileExtension('file')).toEqual('file');
+        expect(removeFileExtension('file.txt')).toEqual('file');
+    });
+
+    it('should get new counter', () => {
+        expect(getNewCounter([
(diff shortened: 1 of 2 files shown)


src/javascript/ContentManager/ContentManager.utils.test.js
```diff
@@ -0,0 +1,27 @@
+import {removeFileExtension, getNewCounter} from './ContentManager.utils';
+
+describe('zip utils', () => {
+    it('should remove file extension', () => {
+        expect(removeFileExtension('file.jpg')).toEqual('file');
+        expect(removeFileExtension('file')).toEqual('file');
+        expect(removeFileExtension('file.txt')).toEqual('file');
```
cedricmoitrier: I don't think this last expectation is adding much value since it's very similar to the first one.
May be valuable to test the following filenames:
- one containing several dots
- one ending with a dot

src/javascript/ContentManager/ContentManager.utils.test.js
```diff
@@ -0,0 +1,27 @@
+import {removeFileExtension, getNewCounter} from './ContentManager.utils';
+
+describe('zip utils', () => {
```
cedricmoitrier: You should use one 'describe' per function to test, with one or many 'it' within.

PR #52, opened 2019-10-28
QA-12302 : adapt CMM to work with any uiLang, as we only support 3 la…

Yousria:
…nguages, now default lang is en and CMM doesn't crash

## JIRA

https://jira.jahia.org/browse/QA-12302

## Description

adapt CMM to have a default UI language if any of our 3 supported languages is set from the user


src/javascript/ContentManager/ContentManager.jsx
```diff
@@ -38,7 +38,13 @@ class ContentManager extends React.Component {
 
         this.defaultNS = 'content-media-manager';
         this.namespaceResolvers = {
-            'content-media-manager': lang => require('../../main/resources/javascript/locales/' + lang + '.json')
+            'content-media-manager': lang => {
+                if (['fr', 'en', 'de'].indexOf(lang) >= 0) {
```
cedricmoitrier: This will prevent customers to provide custom translations. With this we are restricted to only the 3 languages we provide.

cedricmoitrier: This issue applies as well to the changes in other files within this pull request.

PR #59, opened 2019-10-31
QA-12334 : put a condition to avoid access on undefined object

Yousria:
## JIRA

https://jira.jahia.org/browse/QA-12334


src/javascript/ContentManager/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
```diff
@@ -141,7 +141,7 @@ export class FileCard extends Component {
         let rowNumber = isThumbCard ? Math.floor(index / 6) + 1 : Math.floor(index / 2) + 1;
         let columnNumber = isThumbCard ? (index % 6) + 1 : (index % 2) + 1;
         let encodedPath = node.path.replace(/[^/]/g, encodeURIComponent);
-        let isPdf = node.children.nodes.filter(node => node.mimeType.value === 'application/pdf').length !== 0;
+        let isPdf = node.children ? node.children.nodes.filter(node => node.mimeType.value === 'application/pdf').length !== 0 : false;
```
cedricmoitrier: So... no thumbnails for external PDF files. I presume we don't generates any though, right?

PR #210, opened 2020-03-18
BACKLOG-12272

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-12272

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -1,28 +1,44 @@
...
+
+    const renderDefaultContentTrees = registry.add('accordionItem', 'renderDefaultContentTrees', {
+        component: ContentTree,
+        routeComponent: ContentRoute,
+        getPath,
+        getUrlPathPart
     });
 
-    registry.add('accordionItem', JContentConstants.mode.PAGES, renderDefaultContentTrees, {
+    registry.add('accordionItem', 'pages', renderDefaultContentTrees, {
```
DameniMilo: Why don't you keep the constant here?

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -34,11 +50,11 @@ export const jContentAccordionItems = registry => {
         }
     });
 
-    registry.add('accordionItem', JContentConstants.mode.CONTENT_FOLDERS, renderDefaultContentTrees, {
+    registry.add('accordionItem', 'content-folders', renderDefaultContentTrees, {
```
DameniMilo: Why don't you keep the constant here?

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -49,11 +65,11 @@ export const jContentAccordionItems = registry => {
         }
     });
 
-    registry.add('accordionItem', JContentConstants.mode.MEDIA, renderDefaultContentTrees, {
+    registry.add('accordionItem', 'media', renderDefaultContentTrees, {
```
DameniMilo: Why don't you keep the constant here?

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -64,11 +80,18 @@ export const jContentAccordionItems = registry => {
         }
     });
 
-    registry.add('accordionItem', JContentConstants.mode.APPS, {
+    registry.add('accordionItem', 'apps', {
```
DameniMilo: Why don't you keep the constant here?

src/javascript/JContent/JContent.accordion-items.jsx
tdraier: Good question :-) The tab id should be configurable, and we could add as many tabs as we wants - so basically the app name should not be used anywhere else, hence not be a constant (since it's dynamic ...). Unfortunately the constant is still used in different places, and some code is still harcoding usage of some accordions .. but we should get rid of it. I arleady removed  the "apps" constants.

DameniMilo: Ok that's what I thought when seeing the rest of the code.

Good to me then!

PR #215, opened 2020-03-20
[Backlog 12842] Add a CircleCI configuration

rknj:
## Description

With this configuration, a branch is only built if there are PR and, on the master only, the deploy has been added as an additional step

## Checklist

- [x] Performance
- [ ] Migration
- [x] Code maintainability

## Documentation

- [x] Inline documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,3 +42,9 @@ jobs:
       - store_artifacts:
           path: /tmp/artifacts/
       - *persist-workspace
+
+workflows:
+  version: 2
+  on-code-change:
+    jobs:
+      -build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,15 @@ references:
   attach-workspace: &attach-workspace
     attach_workspace:
       at: .
-  
+
 jobs:
+  initialize:
+    <<: *workdir
+    docker:
+      - image: circleci/node:latest
+    steps:
+      - checkout
+
   build:
     <<: *workdir
 
@@ -22,7 +29,7 @@ jobs:
       - image: circleci/openjdk:8-jdk-stretch
 
     steps:
-      - checkout 
+      - *attach-workspace
       - restore_cache:
           keys:
             - v2-dependencies-{{ checksum "pom.xml" }}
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,65 @@
...
+      root: ~/source
+      paths:
+        - .
+
+  attach-workspace: &attach-workspace
+    attach_workspace:
+      at: .
+
+jobs:
+  build:
```
FBerthelot: avoid name a job with the same name of the workflow it can confuse circleci sometimes.
```suggestion
  install:
```

.circleci/config.yml
```diff
@@ -0,0 +1,65 @@
...
+    
+    steps:
+      - *attach-workspace
+      - run: mvn -s .circleci/.circleci.settings.xml deploy
+
+workflows:
+  version: 2
+  build:
+    jobs:
+      - build
```
FBerthelot:
```suggestion
      - install
```

.circleci/config.yml
```diff
@@ -0,0 +1,65 @@
...
+      - deploy:
+          requires:
+            - build
```
FBerthelot:
```suggestion
            - install
```

.circleci/config.yml
Fgerthoffert: You should use context here, this way we can store nexus credentials in one single place.

I just realized this was dropped (probably a mistake on my end) from jahia-ui-root. 
You can specify the context like this:
``` 
- build:
    context: QA_ENVIRONMENT
```   
Documentation is available here: https://circleci.com/docs/2.0/contexts/

I'll add it back to jahia-ui-root, because as-is it's only using the maven cache.

rknj: So I can then remove the settings file ?

Fgerthoffert: you still need the settings file, but the environment variables would be coming from the context instead of the local project env vars

rknj: OK so I just have to remove the variable from the project settings

Fgerthoffert: yes, I'd also clear the maven cache just to be sure settings are properly taken in.

PR #216, opened 2020-03-20
BACKLOG-12877 : Fixed initial loading of additional accordion, fixed …

tdraier:
…loading and warning issue

## JIRA

https://jira.jahia.org/browse/BACKLOG-12877

## Description

- Fixed initial opening of additional tab : takes the first available item, depending on permissions
- Removed state update during render in tree and contentlayout, use effects
- Added a single redux action for changing path/mode/params at the same time to avoid back&forth with the listener leading to inconsistent states
- Removed routes that are not in the current mode, as the url is updated after the mode/path in redux , which lead component trying to render in another mode and executing queries with invalid parameters

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/AdditionalAppsTree/AdditionalAppsTree.jsx
+++ b/src/javascript/JContent/AdditionalAppsTree/AdditionalAppsTree.jsx
@@ -1,26 +1,44 @@
-import React from 'react';
+import React, {useEffect} from 'react';
 import {TreeView} from '@jahia/moonstone';
 import {useDispatch, useSelector} from 'react-redux';
 import {cmGoto} from '../JContent.redux';
 import {useAdminRouteTreeStructure} from '@jahia/ui-extender';
-import {useHistory} from 'react-router';
 import {useNodeInfo} from '@jahia/data-helper';
 import {useTranslation} from 'react-i18next';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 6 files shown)


src/javascript/JContent/AdditionalAppsTree/AdditionalAppsTree.jsx
```diff
@@ -1,26 +1,44 @@
...
+        if (firstItem) {
+            switchSelection = firstItem.key;
+        }
+    }
+
+    useEffect(() => {
+        if (switchSelection) {
+            dispatch(cmGoto({path: '/' + switchSelection}));
+        }
+    }, [dispatch, switchSelection]);
```
cedricmoitrier: What is the reason for passing 'dispatch' here ? I have read https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects and do understand the reason for passing 'switchSelection', but for 'dispatch' I don't get it. Maybe I am missing something...

tdraier: It's quite useless, but since dispatch is used in the body of the callback , it must be part of the dependencies. Removing it won't get through the linter (which will add it automatically, btw)

PR #220, opened 2020-03-24
BACKLOG-12325: Add new breadcrumb (Moonstone based)

cedricmoitrier:
https://jira.jahia.org/browse/BACKLOG-12325

Updates related to selenium tests and removal of "old breadcrumb" will be addressed in dedicated pull requests.

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentBreadcrumb.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentBreadcrumb.jsx
@@ -1,0 +1,16 @@
+import React from 'react';
+
+import styles from './ContentBreadcrumb.scss';
+import ContentPath from './ContentPath';
+import ContentType from './ContentType';
+
+const ContentBreadcrumb = () => {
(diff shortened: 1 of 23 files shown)


tdraier:
- I'm not sure of what we should do with "jmix:visibleInContentTree", but from my understanding it seems we only need to see those items in the breadcrumb (@fpral ?). For now this type is not checked/used anywhere.
- According to the description (is it still accurate ?) : 
> Given I am currently on an area or a content with nested content (like article)
> When I look at the breadcrumb
> Then I see that it starts from the parent page or parent folder
Here it always start from the root, not "from the parent page or parent folder"

cedricmoitrier:
@tdraier I indeed missed that specific point about starting from the parent page/folder for "area/content with nested content".

It seems that the jmix:visibleInContentTree / jmix:cmContentTreeDisplayable part is about identifying what are the parent folders in that case (from my understanding of the Epic description at https://jira.jahia.org/browse/BACKLOG-11608).

But I am still not clear about how to consistently identify an "area" or a "content with nested content" in the first place...

PR #222, opened 2020-03-24
Optimize the CircleCI configuration

rknj:
where the parameter allow to add (or not) the deploy step at the end

## JIRA

https://jira.jahia.org/browse/BACKLOG-12842

## Description

As the "mvn deploy" is also running the "install" tasks, having 2 different jobs do not optimize the execution on master. So the idea here is to use a command with a parameter giving the possibility to add the "deploy" step at the end. Note that I had 2 create 2 jobs using almost the same structure because I did not find how to use the branch as a parameter in a workflow job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ commands:
       - run:
           name: Maven call
           command: |
-          if [[ "${CIRCLE_BRANCH}" == "<< parameters.branch_to_deploy >>" ]]
+          if [[ "${CIRCLE_BRANCH}" == "<< parameters.branch_to_deploy >>" ]];
           then
             mvn -s .circleci/.circleci.settings.xml clean install deploy
           else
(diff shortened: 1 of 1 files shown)


Fgerthoffert:
I feel you could probably simplify a bit the config file using `CIRCLE_BRANCH` (https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-shell-command) instead of having `deploy` and `install_and_deploy` in the workflow section.

The check would happen in `run_maven`. What you could even do is create a parameter (`deploy_branch`) with a default value of master, this way you can extract which branch should be deployed from the core of the config.

Another thing, but we probably can't do anything about it yet, I think circleci recommend to actually break the workflow in multiple jobs, each with a particular focus. But what would compose the actual job is embedded inside the `mvn clean install` so probably best to leave as-is for now but keep this in mind for the future. This would allow to run some actions in parallel.

PR #228, opened 2020-03-26
BACKLOG-12927 : cleanup SearchControlBar and write small test to veri…

Yousria:
…fy multiple selection actions displayed

## JIRA

https://jira.jahia.org/browse/BACKLOG-12927

## Description
cleaned up the SearchControlBar which is part of the Toolbar and it was impossible to write a Jest test with the old version (wrapped with withStyles, connect and so on)
the test is only checking display of selection actions 

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


src/javascript/JContent/ContentRoute/ToolBar/Toolbar.test.jsx
```diff
@@ -0,0 +1,51 @@
...
+jest.mock('connected-react-router', () => jest.fn(() => {}));
+
+describe('Toolbar', () => {
+    it('should not render multiple selection actions', () => {
+        useSelector.mockImplementation(() => ({
+            selection: []
+        }));
+
+        const toolbar = shallow(<ToolBar/>);
+        expect(toolbar.find('Typography').length).toEqual(0);
```
cedricmoitrier: Could with have dedicated tests for the "selection-infos" instead? It's not clear what all those Typography related tests are about while just looking at the tests here.

You could have one test that check that the message is not displayed when there is not selection, and another one testing that it is displayed if there is any selection.

src/javascript/JContent/ContentRoute/ToolBar/Toolbar.test.jsx
```diff
@@ -0,0 +1,51 @@
...
+        expect(toolbar.find('Button').length).toEqual(0);
+    });
+
+    it('should render multiple selection actions for 1 item', () => {
+        useSelector.mockImplementation(() => ({
+            selection: ['/test']
+        }));
+
+        const toolbar = shallow(<ToolBar/>);
+        expect(toolbar.find('Typography').prop('children')).toContain('itemsSelected');
```
cedricmoitrier: I would mock the "translate" function  and check it has been called with the expected parameters (label key and number of items) instead.

src/javascript/JContent/ContentRoute/ToolBar/Toolbar.test.jsx
```diff
@@ -0,0 +1,51 @@
...
+        expect(toolbar.find('Button').prop('icon').type('SvgCancel')).toBeDefined();
+        expect(toolbar.find('Typography').prop('data-cm-selection-size')).toBe(1);
+        expect(toolbar.find('ButtonGroup').prop('children').length).toBe(4);
```
cedricmoitrier: You could use expect(...).toHaveLength(4) to test the length of an array

cedricmoitrier: Actually, I think you could even just write:
expect(toolbar.find('ButtonGroup').children()).toHaveLength(4);

Yousria: ok will use that

src/javascript/JContent/ContentRoute/ToolBar/Toolbar.test.jsx
Yousria: ok, will see how to do that

src/javascript/JContent/ContentRoute/ToolBar/Toolbar.test.jsx
Yousria: sure

src/javascript/JContent/ContentRoute/ToolBar/Toolbar.test.jsx
Yousria: couldn't mock it, as soon as I try to mock react-i18next, it's breaks all my tests, due to some buttons coming from the old design system

cedricmoitrier: Ok, fine. It was just a suggestion.

PR #229, opened 2020-03-26
BACKLOG-12959 : use another target for toolbar and for context menu i…

Yousria:
…n multiple selection

## JIRA

https://jira.jahia.org/browse/BACKLOG-12959

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


tdraier:
Do not add new targets unless there is a very good reason to do it and something completely different (not the case here). Rather filter on the existing ones if you need some actions to be removed. Actually, "selectedContentActions" should also be removed and everything should be based on contentActions (maybe a "support multiple path" option can be added on actions which support multiple selection, to do proper filtering).

PR #238, opened 2020-03-30
BACKLOG-12908: Add selector attributes for Selenium testing

cedricmoitrier:
https://jira.jahia.org/browse/BACKLOG-12908

Add some selector attributes to ContentBreadcrumb for Selenium testing.

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/CompositePathEntry.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/CompositePathEntry.jsx
@@ -18,15 +18,15 @@ const CompositePathEntry = ({items, onItemClick}) => {
     };
 
     return (items.length > 0) &&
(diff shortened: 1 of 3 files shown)


src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/CompositePathEntry.jsx
```diff
@@ -18,14 +18,15 @@ const CompositePathEntry = ({items, onItemClick}) => {
     };
 
     return (items.length > 0) &&
-        <div ref={menuContainerEl}>
```
tdraier: Shouldn't it be "data-sel-role" ?

cedricmoitrier: I used "data-role" everywhere instead, see https://github.com/Jahia/selenium/pull/189
There is no reason to use a naming that references selenium here... "data-role" is more generic.

PR #271, opened 2020-04-08
Implemented conditional rendering of the message based on JContent mode

kovunov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-12809

## Description
Implemented conditional rendering of the message based on JContent mode


tdraier:
Sorry, icon is not there, so looks ok. Still the area looks like a "droppable" area (with dashed borders)

PR #275, opened 2020-04-09
BACKLOG-12677 : avoid rendering progressOverlay for the language and …

Yousria:
…site switcher when loading

## JIRA

https://jira.jahia.org/browse/BACKLOG-12677

## Description

avoid rendering progressOverlay for the language and site switcher when loading

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/LanguageSwitcher.jsx
+++ b/src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/LanguageSwitcher.jsx
@@ -33,7 +33,7 @@ export const LanguageSwitcher = ({
     }
 
     if (loading) {
-        return false;
+        return <Dropdown isDisabled data={[{label: 'none', value: 'none'}]}/>;
     }
 
     return (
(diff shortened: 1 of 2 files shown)


src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/LanguageSwitcher.jsx
```diff
@@ -33,7 +33,7 @@ export const LanguageSwitcher = ({
     }
 
     if (loading) {
-        return <ProgressOverlay/>;
+        return false;
```
tdraier, line 36: The original requirement was to display a disabled dropdown - I don't know if it's possible, but you can check if previous data is available ( siteInfo should contain the previous data ), and use it to display a dropdown , disabled during loading.
If not possible, please update the ticket

Yousria:
Dropdown needs at least one element so I've put a dummy one when loading because the component is doing "const isGrouped = typeof data[0].options !== 'undefined';" which is unsafe maybe we need to improve that, but for now it's working when loading there is disabled dropdowns

PR #291, opened 2020-04-27
QA-12760: use context.id to differentiate upload actions

cedmail:
## JIRA

https://jira.jahia.org/browse/QA-12760

## Description

use context.id to differentiate upload actions, define useEffect dependencies

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


tdraier:
I remember now why we did not destroy the upload component, and so did not use the id - if you destroy the input after the click (so when the upload dialog appear), you cannot upload anything (the input is already destroyed when the user click effectively upload the file)

PR #299, opened 2020-06-11
QA-12920 : use the new field existsInLive to display deletePermanentl…

Yousria:
…y or PublishDeletion actions

## JIRA

https://jira.jahia.org/browse/QA-12920

### DIFF
--- a/src/javascript/JContent/JContent.utils.js
+++ b/src/javascript/JContent/JContent.utils.js
@@ -124,6 +124,19 @@ function uppercaseFirst(string) {
     return string.charAt(0).toUpperCase() + string.substr(1);
 }
 
+const findField = (fieldName, types) => {
+    for (let i = 0; i < types.length; i++) {
+        let type = types[i];
+        for (let j = 0; j < types[i].fields.length; j++) {
+            if (type.fields[j].name === fieldName) {
+                return true;
+            }
+        }
+    }
+
+    return false;
+};
+
 export {
     getNewNodePath,
     hasMixin,
@@ -139,5 +152,6 @@ export {
     removeFileExtension,
     allowDoubleClickNavigation,
     getLanguageLabel,
-    uppercaseFirst
+    uppercaseFirst,
+    findField
 };
--- a/src/javascript/JContent/JContent.utils.js
+++ b/src/javascript/JContent/JContent.utils.js
@@ -124,19 +124,6 @@ function uppercaseFirst(string) {
     return string.charAt(0).toUpperCase() + string.substr(1);
 }
 
-const findField = (fieldName, types) => {
-    for (let i = 0; i < types.length; i++) {
-        let type = types[i];
-        for (let j = 0; j < types[i].fields.length; j++) {
-            if (type.fields[j].name === fieldName) {
-                return true;
-            }
-        }
-    }
-
-    return false;
-};
-
 export {
     getNewNodePath,
     hasMixin,
@@ -152,6 +139,5 @@ export {
     removeFileExtension,
     allowDoubleClickNavigation,
     getLanguageLabel,
-    uppercaseFirst,
-    findField
+    uppercaseFirst
 };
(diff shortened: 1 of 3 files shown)


src/javascript/JContent/actions/deletePermanentlyAction.jsx
```diff
@@ -7,7 +7,8 @@ import {PATH_CONTENTS_ITSELF, PATH_FILES_ITSELF} from './actions.constants';
 
 const checkAction = node => node.operationsSupport.markForDeletion &&
     isMarkedForDeletion(node) &&
-    node.aggregatedPublicationInfo.publicationStatus === 'NOT_PUBLISHED';
+    node.aggregatedPublicationInfo.publicationStatus === 'NOT_PUBLISHED' &&
+    node.aggregatedPublicationInfo.existsInLive !== undefined && !node.aggregatedPublicationInfo.existsInLive;
```
cedricmoitrier, line 11: I don't think this works as expected. As soon as existsInLive is defined then the whole expression will always resolve to false, right?

The green part should be something like:
`publicationStatus === 'NOT_PUBLISHED' && (existsInLive === undefined) ? true : !existsInLive`

I've omitted the 'node.aggregatedPublicationInfo' parts for the sake of clarification here.

src/javascript/JContent/actions/publishDeletionAction.jsx
```diff
@@ -6,7 +6,8 @@ import {useSelector} from 'react-redux';
 
 const checkAction = node => node.operationsSupport.publication &&
     isMarkedForDeletion(node) &&
-    node.aggregatedPublicationInfo.publicationStatus !== 'NOT_PUBLISHED';
+    (node.aggregatedPublicationInfo.publicationStatus !== 'NOT_PUBLISHED' ||
+        (node.aggregatedPublicationInfo.publicationStatus === 'NOT_PUBLISHED' && node.aggregatedPublicationInfo.existsInLive));
```
cedricmoitrier, line 10: You should check whether or not existsInLive is undefined here as well.

src/javascript/JContent/actions/deletePermanentlyAction.jsx
Yousria: you're right we should keep the old behaviour

src/javascript/JContent/actions/publishDeletionAction.jsx
Yousria: done

src/javascript/JContent/JContent.utils.js
```diff
@@ -124,6 +124,19 @@ function uppercaseFirst(string) {
     return string.charAt(0).toUpperCase() + string.substr(1);
 }
 
+const findField = (fieldName, types) => {
```
cedricmoitrier, line 127: I may missing something but how do you know that fieldName exists in the expected type? From my understanding this method will return true if fieldName exists in any type.

cedricmoitrier: Don't we need to either check the type in this method, or parameterize the hook to only return fields from a given type?

src/javascript/JContent/actions/deletePermanentlyAction.jsx
```diff
@@ -1,23 +1,30 @@
...
     isMarkedForDeletion(node) &&
-    node.aggregatedPublicationInfo.publicationStatus === 'NOT_PUBLISHED';
+    node.aggregatedPublicationInfo.publicationStatus === 'NOT_PUBLISHED' &&
+    (node.aggregatedPublicationInfo.existsInLive === undefined ? true : !node.aggregatedPublicationInfo.existsInLive);
 
 export const DeletePermanentlyActionComponent = ({context, render: Render, loading: Loading}) => {
     const {language} = useSelector(state => ({language: state.language}));
 
+    let types = useSchemaFields();
+    let existsInLiveField = types === undefined ? false : findField('existsInLive', types);
```
tdraier, line 19: This code should be done directly by useNodeInfo

src/javascript/JContent/JContent.utils.js
tdraier: yes, agree - you should always ask for a field in a specific type , not look in all types

src/javascript/JContent/actions/deletePermanentlyAction.jsx
Yousria: Ok

PR #310, opened 2020-07-17
BACKLOG-14186 add breadcrumb in content editor

Taybou:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14186

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
@@ -1,10 +1,12 @@
 import React, {useMemo} from 'react';
+import PropTypes from 'prop-types';
 import {useDispatch, useSelector} from 'react-redux';
 import {useQuery} from '@apollo/react-hooks';
 
 import {cmGoto} from '~/JContent/JContent.redux';
 import {GetContentPath} from './ContentPath.gql-queries';
 import ContentPath from './ContentPath';
+import JContentConstants from '../../../JContent.constants';
 
 function findLastIndex(array, callback) {
     let lastIndex = -1;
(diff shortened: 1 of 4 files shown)


dgriffon (commented):
Globally it seems odd for jContent to expose the breadcrumb:
- There is no usage of that code in jContent, making it difficult to understand. 
- it adds complexity to the component that would have been avoid if the CE logic had been put in its own implementation. 

If we choose to keep that implementation:
- jContent should also use the registered component. 
- the component should be more generic and take a path as parameter

tdraier (changes requested):
You cannot use the regsitry to expose a component. Share the different parts in external libraries or recopy the code.
If you really need to share the logic, create a hook in data-helper (as we did for preview).

src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
```diff
@@ -1,10 +1,12 @@
 import React, {useMemo} from 'react';
+import PropTypes from 'prop-types';
 import {useDispatch, useSelector} from 'react-redux';
 import {useQuery} from '@apollo/react-hooks';
 
 import {cmGoto} from '~/JContent/JContent.redux';
 import {GetContentPath} from './ContentPath.gql-queries';
 import ContentPath from './ContentPath';
+import JContentConstants from '../../../JContent.constants';
```
cedricmoitrier, line 9:
```suggestion
import JContentConstants from '~/JContent/JContent.constants';
```

src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.container.jsx
```diff
@@ -47,20 +49,34 @@ const ContentPathContainer = () => {
...
+        }));
     };
 
     if (error) {
         console.log(error);
     }
 
     const node = data?.jcr?.node || {};
-    const items = useMemo(() => getItems(node), [node]);
+    const items = useMemo(() => getItems(externalPath, node), [node, externalPath]);
```
cedricmoitrier, line 70: You should use the node returned by useQuery here... I don't understand the need for passing externalPath.

cedricmoitrier (changes requested):

Taybou:
Thank you for the review, I'll close this PR.

PR #313, opened 2020-07-27
BACKLOG-14246 update breadcurmb separator style

Taybou:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14246

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import {Breadcrumb} from '@jahia/moonstone';
+import {Breadcrumb, Separator} from '@jahia/moonstone';
 
 import CompositePathEntry from './CompositePathEntry';
 import SimplePathEntry from './SimplePathEntry';
@@ -20,10 +20,12 @@ const renderItems = (items, onItemClick) => {
 
 const ContentPath = ({items, onItemClick}) => {
     return (items.length > 0) &&
-        <Breadcrumb>
-            {renderItems(items, onItemClick)}
-            <></>
-        </Breadcrumb>;
+        <>
+            <Breadcrumb>
+                {renderItems(items, onItemClick)}
+            </Breadcrumb>
+            <Separator variant="vertical"/>
+        </>;
 };
 
 ContentPath.propTypes = {


src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.jsx
```diff
@@ -20,10 +20,12 @@ const renderItems = (items, onItemClick) => {
...
     return (items.length > 0) &&
-        <Breadcrumb>
-            {renderItems(items, onItemClick)}
-            <></>
-        </Breadcrumb>;
+        <>
+            <Breadcrumb>
+                {renderItems(items, onItemClick)}
+            </Breadcrumb>
+            <Separator variant="vertical"/>
```
cedricmoitrier, line 27: Will this separator be properly styled as the ones inside the Moonstone's Breadcrumb without any additional changes?

Taybou: yes, as recommended by the UX

cedricmoitrier, line 27: It should be moved ContentBreadcumb.jsx instead (between ContentPath and ContentType components).

cedricmoitrier (changes requested):

src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.jsx
Taybou: I don't understand? What do you mean?

cedricmoitrier: The "Separator" component should be moved out of ContentPath. It does not make sense here since we are no longer leveraging the one from the Breadcrumb. Proper location should be one level above - between <ContentPath/> and <ContentType/> components (ie. in ContentBreadcrumb.jsx).

cedricmoitrier: here: https://github.com/Jahia/jcontent/blob/master/src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentBreadcrumb.jsx

Taybou: I finally see your point of view, from the way I put it here, to avoid in case we don't have a content path (items.length === 0) so we'll have an ugly pipe (separator) and a content type; this particular case can happen if we are at the root level of the content ex: /contents or /home

cedricmoitrier: Fair point. Thanks for clarifying!

cedricmoitrier (approved):

src/javascript/JContent/ContentRoute/ContentBreadcrumb/ContentPath/ContentPath.jsx
Taybou: you're welcome

PR #317, opened 2020-07-29
BACKLOG-14015 : use props instead of context object

Yousria:
## JIRA

https://jira.jahia.org/browse/BACKLOG-14015

### DIFF
--- a/src/javascript/JContent/ContentRoute/ToolBar/ToolBar.jsx
+++ b/src/javascript/JContent/ContentRoute/ToolBar/ToolBar.jsx
@@ -58,7 +58,9 @@ export const ToolBar = () => {
                         target="selectedContentActions"
                         context={context}
                         filter={action => action.key !== 'deletePermanently' && action.key.indexOf('publish') === -1}
-                        render={getButtonRenderer({size: 'default', variant: 'ghost'})}
+                        size="default"
+                        variant="ghost"
+                        render={getButtonRenderer()}
                     />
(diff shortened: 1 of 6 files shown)


src/javascript/JContent/ContentRoute/ToolBar/ToolBar.jsx
```diff
@@ -58,7 +58,9 @@ export const ToolBar = () => {
                         target="selectedContentActions"
                         context={context}
                         filter={action => action.key !== 'deletePermanently' && action.key.indexOf('publish') === -1}
-                        render={getButtonRenderer({size: 'default', variant: 'ghost'})}
+                        size="default"
```
cedricmoitrier, line 61: Shouldn't 'size', 'variant' and 'isReversed' remains props of the ButtonRenderer instead of DisplayAction/DisplayActions... since those are specific to the Button component?

tdraier: good question, but it was actually the original requirement to simplify the way to pass props to the different components in the chain - display action, action component, renderer and finally the button itself (or whatever is renderer), without the need to create specific code for that (like in getButtonRenderer).

tdraier: note that they could still be wrapped in a "buttonProps" prop to avoid passing all props into the button

PR #325, opened 2020-08-10
QA-13058 Provide 404 message when app is not available, check module …

AKarmanov:
…availability on site

## JIRA

https://jira.jahia.org/browse/QA-13058

## Description

Show message when app is not available and filter out routes based on modules installed on site

### DIFF
--- a/src/javascript/JContent/AdditionalAppsRoute/AdditionalAppsRoute.jsx
+++ b/src/javascript/JContent/AdditionalAppsRoute/AdditionalAppsRoute.jsx
@@ -4,22 +4,34 @@ import {useNodeInfo} from '@jahia/data-helper';
 import {Route} from 'react-router';
 import React from 'react';
 import PropTypes from 'prop-types';
+import {Typography} from '@jahia/moonstone';
+import {useTranslation} from 'react-i18next';
+import styles from './AdditionAppsRoute.scss';
 
 export const AdditionalAppsRoute = ({match, target}) => {
     const site = useSelector(state => state.site);
+    const {t} = useTranslation('jcontent');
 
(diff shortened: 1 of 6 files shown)


src/javascript/JContent/AdditionalAppsRoute/AdditionalAppsRoute.jsx
```diff
@@ -4,22 +4,34 @@ import {useNodeInfo} from '@jahia/data-helper';
...
-        .filter(route => route.isSelectable && route.render);
+        .filter(route => route.isSelectable && route.render)
+        .filter(route => route.requireModuleInstalledOnSite === undefined || (node && node.site.installedModulesWithAllDependencies.indexOf(route.requireModuleInstalledOnSite) !== -1));
 
     return (
         <>
             {filteredAdminRoutes && filteredAdminRoutes.map(r =>
                 <Route key={r.key} path={match.path + '/' + r.key} render={props => r.render(props)}/>
             )}
+            <Route key="nothingToDisplayRoute"
```
tdraier, line 31: This route will render in all cases, it should be wrapped in a <Switch> if we only want one to be used

AKarmanov: I'll check, but I thought that the routes returned were already part of a switch, maybe I didn't notice it because it rendered at the bottom. Anyways I'll check.

AKarmanov: You were right, it was rendering, there is no switch wrapping it. And it didn't show because iframe was covering it.

tdraier (approved):

PR #331, opened 2020-08-21
QA-13098 Fix multiple selection issue by adding a check for undefined

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-13098

## Description

Fix context.pages undefined issue when selecting multiple files

### DIFF
--- a/src/javascript/JContent/actions/publishAction.jsx
+++ b/src/javascript/JContent/actions/publishAction.jsx
@@ -62,7 +62,7 @@ function getButtonLabelParams(context, language, res, t) {
     }
 
     return {
-        displayName: t('jcontent:label.contentManager.selection.items', {count: context.pages ? context.pages.length : 0}),
+        displayName: t('jcontent:label.contentManager.selection.items', {count: context.paths}),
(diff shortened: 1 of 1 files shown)


src/javascript/JContent/actions/publishAction.jsx
```diff
@@ -62,7 +62,7 @@ function getButtonLabelParams(context, language, res, t) {
     }
 
     return {
-        displayName: t('jcontent:label.contentManager.selection.items', {count: context.pages.length}),
+        displayName: t('jcontent:label.contentManager.selection.items', {count: context.pages ? context.pages.length : 0}),
```
tdraier, line 65: What is context.pages ? Something happened at some point, but for me it should be context.paths

tdraier (changes requested):

src/javascript/JContent/actions/publishAction.jsx
AKarmanov: Ahh, you are right of course, it was introduced by me when I moved it, not sue how I went from paths to pages given that I copied the code. My only explanation is that I work on router issue in parallel which got me thinking about pages.

tdraier (approved):

PR #332, opened 2020-09-08
QA-13071

tdraier:
## JIRA

https://jira.jahia.org/browse/QA-13071

## Description

This fix removes the exception and gives access to cmm, but you will still have an error when trying to load the language switcher (which does not appear)

### DIFF
--- a/src/main/java/org/jahia/modules/contentmanager/configuration/URLResolverFactory.java
+++ b/src/main/java/org/jahia/modules/contentmanager/configuration/URLResolverFactory.java
@@ -49,6 +49,7 @@ import org.jahia.api.Constants;
 import org.jahia.exceptions.JahiaRuntimeException;
 import org.jahia.registries.ServicesRegistry;
 import org.jahia.services.cache.ehcache.EhCacheProvider;
+import org.jahia.services.content.JCRSessionFactory;
 import org.jahia.services.render.RenderContext;
 import org.jahia.services.sites.JahiaSite;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jahia/modules/contentmanager/configuration/URLResolverFactory.java
```diff
@@ -102,6 +103,9 @@ private String convertPathinfo(String pathInfo) {
         try {
             servlet =  pathParts[CM_SERVLET_IDX];
             JahiaSite jahiaSite = ServicesRegistry.getInstance().getJahiaSitesService().getSiteByKey(pathParts[SITE_KEY_IDX]);
+            if (jahiaSite == null) {
+                jahiaSite = ServicesRegistry.getInstance().getJahiaSitesService().getSiteByKey(pathParts[SITE_KEY_IDX], JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE));
```
cedricmoitrier, line 107: Can't we just replace the first call (at line 105) by this one then?

cedricmoitrier (approved):

src/main/java/org/jahia/modules/contentmanager/configuration/URLResolverFactory.java
tdraier: yes, actually users should have access to the site in edit anyway.

cedricmoitrier (approved):

PR #342, opened 2020-10-23
QA-13349 : avoid blank page when rename content from content editor

dgriffon:
https://jira.jahia.org/browse/QA-13349

### DIFF
--- a/src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/SiteSwitcher.jsx
+++ b/src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/SiteSwitcher.jsx
@@ -95,7 +95,7 @@ class SiteSwitcher extends React.Component {
                         return null;
                     }
 
-                    if (loading) {
+                    if (loading || (data && data.jcr === null)) {
                         return (
                             <Dropdown isDisabled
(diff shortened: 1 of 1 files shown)


jsinovassin (approved):

src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/SiteSwitcher.jsx
```diff
@@ -95,7 +95,7 @@ class SiteSwitcher extends React.Component {
                         return null;
                     }
 
-                    if (loading) {
+                    if (loading || (data && data.jcr === null)) {
```
cedricmoitrier, line 98: I am a bit confuse by this change. What method is failing exactly? Isn't it the call to getSites(data). If that is so, what about changing line 51 (in getSites() function) instead to check for data.jcr not being null?

dgriffon: This was a workaround until we found the reason of the issue,
It was added as the loading behaviour because it should display the same empty component when it happens.

dgriffon: The main issue has been fixed

PR #351, opened 2020-11-20
BACKLOG-13977: Displaying the weight of the image

nbugash-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13977

## Description

When viewing an image, aside from seeing the height and width, you should now be able to see the size (or weight) of the image.

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentLayout/ContentLayout.gql-queries.js
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/ContentLayout.gql-queries.js
@@ -183,6 +183,9 @@ class FilesQueryHandler {
                                 }
                                 children(typesFilter: {types: ["jnt:resource"]}) {
                                     nodes {
+                                        data: property(name: "jcr:data") {
+                                            value
+                                        }
                                         mimeType: property(name: "jcr:mimeType") {
                                             value
                                         }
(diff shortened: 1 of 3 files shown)


src/javascript/JContent/ContentRoute/ContentLayout/ContentLayout.gql-queries.js
```diff
@@ -183,6 +183,9 @@ class FilesQueryHandler {
                                 }
                                 children(typesFilter: {types: ["jnt:resource"]}) {
                                     nodes {
+                                        data: property(name: "jcr:data") {
```
cedricmoitrier, line 186: Looks like you are retrieving the whole binary. That will likely have a huge impact on performance right?
The length should likely be computed on the server side instead (if not stored anywhere).

cedricmoitrier (changes requested):

src/javascript/JContent/ContentRoute/ContentLayout/ContentLayout.gql-queries.js
nbugash-jahia: ok.i'll take a look

nbugash-jahia:
I'm closing this PR and going to create a new one since there since there some changes made to the app-shell module and tbh, I don't want to deal with rebasing the branch (cause a lot of headaches).

PR #361, opened 2020-12-10
BACKLOG-15148: Fix search results no longer being displayed

cedricmoitrier:
## JIRA
https://jira.jahia.org/browse/BACKLOG-15148

## Description
Regression has been introduced while fixing BACKLOG-15053


gflores-jahia (approved):
Looks good to me code-wise. Seems to be checking more generically on result end nodes than specific path.

PR #368, opened 2020-12-17
BACKLOG-13979: fixing card file image size and colour.

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13979

## Description

Very simple CSS change to the FileCard to display the background image as "contain" instead of "cover" as the size of the image is always greater than height 200px width 200px (due to the image we are using)
Very simple CSS addition to colour the background colour to #DADADA as per the Figma documentation.


gflores-jahia (approved):
awesome concise fix! lgtm

nbugash-jahia (approved):
lgtm

PR #371, opened 2020-12-23
BACKLOG-13981: Add-in the multi-select feature to grid view.

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13981

## Description

Relatively simple PR to use the existing selection redux state for gird view as well.
There is special request on initial if not 'checked' to not display the checkbox only when hovered; but once checked the will remain displayed unless 'unchecked'.
The shadowbox is slightly different from the existing used for preview as well (noted to @fpral )

## Testing

Please see below video for the unit testing
https://user-images.githubusercontent.com/13266177/103023829-aa178100-451c-11eb-8733-ef08617a2f52.mp4


gflores-jahia (approved):

gflores-jahia:
not technically a unit test but better than nothing.  at some point might need to setup a separate story to at least start putting in test framework for jcontent.

PR #372, opened 2020-12-23
BACKLOG-13995: Adding the number of subnode per each folder.

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13995

## Description

Simple PR to port the folder sub-node badge feature to grid view as well.

NOTE: on testing found that on 'upload' action the registry state is not refreshed. (might be a separate concern)

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
@@ -224,8 +224,7 @@ export const FileCard = ({
                             {node?.subNodes?.pageInfo?.totalCount}
                             &nbsp;
                             {node?.subNodes?.pageInfo?.totalCount === 1 ?
-                                t('jcontent:label.contentManager.filesGrid.element')
-                                : t('jcontent:label.contentManager.filesGrid.elements')}
+                                t('jcontent:label.contentManager.filesGrid.element') : t('jcontent:label.contentManager.filesGrid.elements')}
                         </Typography>
                     </div>}
                 </CardContent>
--- a/src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
@@ -144,6 +144,7 @@ export const FileCard = ({
     let encodedPath = node.path.replace(/[^/]/g, encodeURIComponent);
     let isPdf = node.children ? node.children.nodes.filter(node => node.mimeType.value === 'application/pdf').length !== 0 : false;
     let isCardSelected = selection ? selection.includes(node.path) : false;
+    let showSubNodes = node.primaryNodeType.name !== 'jnt:page' && node?.subNodes?.pageInfo?.totalCount > 0;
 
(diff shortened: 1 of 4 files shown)


src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
```diff
@@ -117,6 +124,9 @@ export const FileCard = ({
     let columnNumber = (index % 2) + 1;
     let encodedPath = node.path.replace(/[^/]/g, encodeURIComponent);
     let isPdf = node.children ? node.children.nodes.filter(node => node.mimeType.value === 'application/pdf').length !== 0 : false;
+
+    let showSubNodes = node.primaryNodeType.name !== 'jnt:page' && node.subNodes && node.subNodes.pageInfo.totalCount > 0;
```
gflores-jahia, line 128: do we need to add null check for pageInfo in there? alternatively might be better to use optional chaining e.g. `node?.subNodes?.pageInfo?.totalCount > 0`. Also make sure it's returning boolean and not falsey value as falsey might still render

src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
```diff
@@ -185,6 +195,14 @@ export const FileCard = ({
                             {`${node.width.value} x ${node.height.value}`} <FileSize node={node}/>
                         </Typography>
                     </div>}
+                    {showSubNodes &&
```
gflores-jahia, line 198: See showSubNodes comment above. Need to make sure it's a boolean and not just falsey value.

src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
XiangYuHe: Oh cool never knew that awesome

gflores-jahia: if showSubNodes is null or undefined then I believe it's fine. but for some other falsey values (e.g. 0 or Nan), those values can still get rendered https://dev.to/aviskarkc10/jsx-gotcha-when-conditionally-rendering-using-179j

In general, I'd just try to avoid falsey values and just convert to straight up boolean (using !!) or ternary operator.

XiangYuHe: Gotcha the example make sense
Kindly take another look 
As with the optional chaining it will return undefined if the value does not exist (which is good in our case)
Plus other values will be either true or false (believe we are covered)

src/javascript/JContent/ContentRoute/ContentLayout/FilesGrid/FileCard/FileCard.jsx
XiangYuHe: Yup I believe with switching to optional chaining it will be a boolean value or null/undefined (in both cases it should be OK)

gflores-jahia (approved):

gflores-jahia (approved):
looks like commits were just squished. lgtm

PR #376, opened 2021-01-05
BACKLOG-15198 used hack with space in from condition to fix sql2 search

kovunov:
## JIRA
https://jira.jahia.org/browse/BACKLOG-15198

## Description
Used space addition to sql2 FROM parameter, which fixes this issue and doesn't break existing functionality, though it's hacky I don't see the reason why encoding/decoding without space doesn't work

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@material-ui/core": "^3.6.2",
     "@material-ui/icons": "^3.0.2",
     "@material-ui/lab": "^3.0.0-alpha.16 ",
+    "base64url": "^3.0.1",
     "bytes": "^3.1.0",
     "classnames": "^2.2.5",
     "connected-react-router": "^4.4.1",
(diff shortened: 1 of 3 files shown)


tdraier (changes requested):
Unfortunately the "hack" is not valid - it fixes this specific string, but will not fix all queries. Basically changing any character in the param fixes the issue , meaning changing another character can lead to the same issue.
What I've found is that btoa is creating base64 string with + in it, which is translated by queryString.parse into ' ' , and so impossible to decode with a atob afterwards.
Base64 is not url safe, 
A standard "url safe base64" way is to replace `+` by `-`, `/` by `_` , and remove the `=` padding (not needed to decode). For example check the encoding of the following string : 
```(sql2SearchFrom:bigText,sql2SearchWhere:'[text] like !'%Digitall%!' AND [jcr:created] > cast(!'2016-01-10T00:00:00.000+02:00!' as date)')```
In https://www.base64url.com/

tdraier (approved):

PR #380, opened 2021-01-18
BACKLOG-15355 Propagate className prop to moonstone button

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15355

## Description

Use className prop with moonstone button

### DIFF
--- a/src/javascript/utils/getButtonRenderer.jsx
+++ b/src/javascript/utils/getButtonRenderer.jsx
@@ -5,7 +5,7 @@ import React from 'react';
 
 const getButtonRenderer = ({labelStyle} = {}) => {
     const ButtonRenderer = props => {
-        const {buttonLabelNamespace, buttonLabelShort, buttonLabel, isVisible, buttonLabelParams, buttonIcon, actionKey, enabled, disabled, onClick, buttonProps} = props;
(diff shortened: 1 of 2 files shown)


src/javascript/utils/getButtonRenderer.jsx
```diff
@@ -5,7 +5,7 @@ import React from 'react';
 
 const getButtonRenderer = ({labelStyle} = {}) => {
     const ButtonRenderer = props => {
-        const {buttonLabelNamespace, buttonLabelShort, buttonLabel, isVisible, buttonLabelParams, buttonIcon, actionKey, enabled, disabled, onClick, buttonProps} = props;
+        const {buttonLabelNamespace, buttonLabelShort, buttonLabel, isVisible, buttonLabelParams, buttonIcon, actionKey, enabled, disabled, onClick, buttonProps, className} = props;
```
tdraier, line 8: why not using the buttonProps ?

AKarmanov: It looks like it worked with className on DisplayAction at some point (unless it was totally missed the first time), I didn't want to introduce something that would require a larger change. I'll check how many of these display actions with className there are. I do agree that button props would be a better place for it.

tdraier (approved):

src/javascript/utils/getButtonRenderer.jsx
tdraier: ok i don't know if it worked before and why, if it was used like that and there's many places to update, using className prop directly is ok

PR #381, opened 2021-02-04
QA-13488: Allowing the visibility of the cut to not be restricted like COPY

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/QA-13488

## Description

Confirmation with @fpral to remove the constraints on CUT to display for "Pages" as well.

**NOTE**: one interesting note is both showOnNodeTypes and hideOnNodeTypes filter the existing items based on the logic not sure if it is mean to be this way.

### DIFF
--- a/src/javascript/JContent/actions/actions.constants.js
+++ b/src/javascript/JContent/actions/actions.constants.js
@@ -1,4 +1,2 @@
 export const PATH_CONTENTS_ITSELF = '^/sites/((?!/).)+/contents/?$';
-export const PATH_CONTENTS_DESCENDANTS = '^/sites/((?!/).)+/contents/.+';
 export const PATH_FILES_ITSELF = '^/sites/((?!/).)+/files/?$';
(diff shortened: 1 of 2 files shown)


gflores-jahia (approved):

gflores-jahia:
Clarification: are we showing visibility of both cut and copy now, since we've removed all restrictions? And is this similar behavior to the backported change (I still see hide restriction on the 7.3.8.x change)

XiangYuHe:
> Clarification: are we showing visibility of both cut and copy now, since we've removed all restrictions? And is this similar behavior to the backported change (I still see hide restriction on the 7.3.8.x change)

Right we are showing the visibility of both cut and copy instead of just copy.
I am trying to keep the back-port consistent with v8 that is why I added. 
Being said I took an extra look at useNodeChecks.js the showOnNodeTypes seems like is filtering those variable out... (That is why) bad naming perhaps.

gflores-jahia (approved):

tdraier (approved):

PR #384, opened 2021-02-16
BACKLOG-15325: refactor the tag to use registry from automated-tags

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15325

## Description

Moving the set-up of code of windows to automated-tags repo.
Getting the mutation from registry.


src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
```diff
@@ -189,37 +190,46 @@ export class UploadItem extends React.Component {
         }
 
         if (type === 'replace') {
+            if (window.contextJsParameters.config.links.tagImageOnUpload) {
```
tdraier, line 193: Using the registry can be ok, but it must be generic and not introduce a dependency to automated tag module. You can simple use a registry.find({type:'fileUpload'}), iterate on the results and take the first callback that can handle he upload. For more clarity I would replace all 4 client.mutate here by registry entries, and completely replace the handleUpload method by just :   
```
    handleUpload(type) {
        return registry.find({type: 'fileUpload'}).find(upload => upload.canHandle(type, this.props)).handleUpload(type, this.props);
    }
```

The "default" fileupload handlers would be registered by jcontent initialization by something like : 
```
        registry.add('fileUpload', 'import', {
            canHandle: type => type === 'import',
            handleUpload: (type, {path, file, client}) => (
                client.mutate({
                    mutation: importContent,
                    variables: {
                        path: path,
                        fileHandle: file
                    }
                })
            )
        });
```

I did not check the previous PR, but all reference to AutomatedTags must be removed from jcontent.

tdraier (changes requested):

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
XiangYuHe: Thanks @tdraier I see what you mean (appreciated for the code snippet)

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
```diff
@@ -143,9 +141,9 @@ export class UploadItem extends React.Component {
             setTimeout(() => {
                 this.props.uploadFile(upload);
                 this.props.updateUploadsStatus();
-                const tagsList = updateReturnObj?.data?.jcr?.addNode?.awsRecognition?.tagImageSync;
-                if (typeof tagsList !== 'undefined' && tagsList !== null) {
-                    this.setState({numTags: tagsList.length});
+                const tagsComponent = uploadReturnObj;
```
tdraier, line 144: maybe rather use a property of "uploadReturnObj" (`uploadReturnObj.component`), if we need to pass other data later.

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
```diff
@@ -179,49 +177,20 @@ export class UploadItem extends React.Component {
...
-        return client.mutate({
-            mutation: uploadFile,
-            variables: {
-                fileHandle: file,
-                nameInJCR: this.getFileName(),
-                path: path,
-                mimeType: file.type,
-                isImage: isImageFile(file.name)
-            }
-        });
```
tdraier, line 224: As proposed earlier use something more generic, and add the `canHandle` into registry entries :  
```suggestion
        return registry.find({type: 'fileUpload'}).find(upload => upload.canHandle(type, this.props)).handleUpload(type, this.props);
```

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
```diff
@@ -61,7 +59,7 @@ export class UploadItem extends React.Component {
         this.state = {
             userChosenName: null,
             anchorEl: null,
-            numTags: '-'
+            tagComp: ''
```
tdraier, line 62: just name component to be more generic, rather set it to null as default to keep type consistent

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
tdraier: you can also make it even more simple by using : 
```return registry.get('fileUpload', type).handleUpload(this.props);``` , where the name of your registry entry would match the type

XiangYuHe: I actually checked this portion to see if I can re-factor this into the one you are currently specifying.
There are two limiting factors:
1. For regular upload the **type** value is actually "undefined" (I don't believe it is a good idea to register a type of "undefined"
2. Existing code has this functionality to get specific filename on "replace" and "undefined" based on state variables in `this.getFileName()`. Props doesn't have this value so need some special handling.

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
XiangYuHe: Sounds good updated

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
XiangYuHe: Yup OK sounds good Updated

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.jsx
tdraier: I think that's ok for now, we may want to improve that later. Quite a shame we have this `getFilename()` stuff here.

tdraier (approved):

PR #385, opened 2021-02-22
BACKLOG-15625 issue with z-index on page composer add in return of uuid.

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15625

## Description

Issue with "page composer" as it has display issue with z-index in comparison with jcontent.
Business requested that the "edit" onclick to open in a new tab instead.

The reason I am not using the existing code is the action is tightly bind with redux action.
Plus do not want to affect existing "edit" functionalities.

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/EditButton.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/EditButton.jsx
@@ -1,25 +1,39 @@
 import React from 'react';
 import {uploadStatuses} from '../Upload.constants';
-import {DisplayActions} from '@jahia/ui-extender';
-import {ButtonRenderer} from '~/utils/getButtonRenderer';
 import {OpenInNew} from '@jahia/moonstone/dist/icons';
+import {Button} from '@jahia/moonstone';
 import PropTypes from 'prop-types';
+import {useSelector} from 'react-redux';
 import {isImageFile} from '../../ContentLayout.utils';
 
+const showEditButton = ({uuid, status}) => {
+    if (uuid !== null && status === uploadStatuses.UPLOADED) {
+        return true;
+    }
+
+    return false;
+};
+
 const EditButton = props => {
-    const {status, file, t, path} = props;
+    const {status, file, t, uuid} = props;
+
+    const {language} = useSelector(state => ({language: state.language}));
+    const windowLocationPathNames = window.location.pathname.split('/');
+    const location = (windowLocationPathNames.length > 1) ? `/${windowLocationPathNames[1]}` : '';
+    const url = `${location}/content-editor/${language}/edit/${uuid}`;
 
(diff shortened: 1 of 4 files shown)


src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/EditButton.jsx
```diff
@@ -1,25 +1,30 @@
...
+import {useSelector} from 'react-redux';
 import {isImageFile} from '../../ContentLayout.utils';
 
 const EditButton = props => {
-    const {status, file, t, path} = props;
+    const {status, file, t, uuid} = props;
 
-    if (isImageFile(file.name)) {
+    const {language} = useSelector(state => ({language: state.language}));
+    const location = (window.location.pathname.split('/').length > 1) ? `/${window.location.pathname.split('/')[1]}` : '/';
```
AKarmanov: Maybe we can extract window.location.pathname.split('/') to constant so we don't do it twice and it's a little bit shorter?

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/EditButton.jsx
```diff
@@ -1,25 +1,30 @@
...
+    const url = `${location}/content-editor/${language}/edit/${uuid}`;
+
+    if (isImageFile(file.name) && uuid !== null) {
```
AKarmanov, line 16: Can these two ifs bee merged into one?

XiangYuHe: Personally I would put those differently 
As one is checking the uuid of the node it self while the other is checking if file is an image (different purpose)

AKarmanov: For me it's more about not having complexity of nested IF, in the end the purpose of the if check is to determine if component should be displayed so I think we would be totally ok merging the two.

XiangYuHe: Ah I see where you are going with that
I can combo those into an additional function that call displayEditButton

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/EditButton.jsx
XiangYuHe: Yup sounds good.

AKarmanov (approved):

src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/EditButton.jsx
XiangYuHe: Updated to 1 if statement 👍

PR #390, opened 2021-03-15
QA-13567 use the sortBy introduced to sort the useTreeEntries results.

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/QA-13567

## Description

Using the sort by to get the sorted results

Example:
`query test($path: String!, $sorter: InputFieldSorterInput) {
    jcr {
      nodeByPath(path: $path) {
        children(fieldSorter: $sorter) {
          nodes {
            name
          }
        }
      }
    }
}`

`{
  "path":"/sites/digitall/files",
  "sorter":{ "fieldName": "name", "sortType": "ASC"}
}`

### DIFF
--- a/src/javascript/JContent/ContentTree/ContentTree.jsx
+++ b/src/javascript/JContent/ContentTree/ContentTree.jsx
@@ -18,7 +18,7 @@ export const ContentTree = ({lang, siteKey, path, openPaths, setPath, openPath, 
         openPaths.push(rootPath);
     }
 
-    const json = {
+    const useTreeEntriesOptionsJson = {
         fragments: [PickerItemsFragment.mixinTypes, PickerItemsFragment.primaryNodeType, PickerItemsFragment.isPublished, lockInfo, displayName],
         rootPaths: [rootPath],
         openPaths: openPaths,
@@ -31,10 +31,10 @@ export const ContentTree = ({lang, siteKey, path, openPaths, setPath, openPath, 
 
     // For pages portion want to skip the sortBy
     if (item.key !== 'pages') {
-        json.sortBy = SORT_CONTENT_TREE_BY_NAME_ASC;
+        useTreeEntriesOptionsJson.sortBy = SORT_CONTENT_TREE_BY_NAME_ASC;
     }
 
-    const {treeEntries, refetch} = useTreeEntries(json);
+    const {treeEntries, refetch} = useTreeEntries(useTreeEntriesOptionsJson);
 
     let switchPath;
     // If path is root one but root is hidden, then select its first child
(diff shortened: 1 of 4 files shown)


src/javascript/JContent/ContentTree/ContentTree.jsx
```diff
@@ -25,7 +26,8 @@ export const ContentTree = ({lang, siteKey, path, openPaths, setPath, openPath,
         openableTypes: item.config.openableTypes,
         selectableTypes: item.config.selectableTypes,
         queryVariables: {language: lang},
-        hideRoot: item.config.hideRoot
+        hideRoot: item.config.hideRoot,
+        sortBy: SORT_CONTENT_TREE_BY_NAME_ASC
```
AKarmanov, line 30: Does this sort pages tree as well and if it does do we want to do this? We also need to have yarn.lock updated with updated version of dependencies.

XiangYuHe: Yes it sorts the pages tree as well
I will check with @fpral on that one
Thanks (I will note it to update the yarn.lock with the new dependency) good catch

XiangYuHe: Updated the package.json and yarn.lock
Waiting on feedback from @fpral

XiangYuHe: Feed-back to update this portion to exclude for item.key === pages

src/javascript/JContent/ContentTree/ContentTree.jsx
```diff
@@ -17,7 +18,7 @@ export const ContentTree = ({lang, siteKey, path, openPaths, setPath, openPath,
         openPaths.push(rootPath);
     }
 
-    const {treeEntries, refetch} = useTreeEntries({
+    const json = {
```
AKarmanov, line 21: I would probably call it something other than json :) but other than that looks fine

AKarmanov (approved):

src/javascript/JContent/ContentTree/ContentTree.jsx
XiangYuHe: The universal json variable is all I can think about haha
Maybe `useTreeOptionsJson` is better?

AKarmanov: Anything that would tie it to the fact that it is used in the useTreeEntries hook is better for me )

XiangYuHe: okie sounds good

PR #394, opened 2021-03-26
TECH-400: Use error boundaries and loader

tdraier:
## JIRA

https://jira.jahia.org/browse/TECH-400


vindhya (approved):

vindhya:
Looks good! I just wasn't able to force an error to test out the error boundary.

PR #401, opened 2021-04-23
BACKLOG-15824: permissions on jcontent accordions

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15824

## Description

Initial display of accordions based on permissions.
Then if permission is missing then on click on the jContent button default to the proper URL.

- [x] In-case of no permission what is the display (confirmed to show at /<lang> blank page)
- [x] Custom permission not really dynamic in a sense (confirmed)

### DIFF
--- a/src/javascript/JContent.register.jsx
+++ b/src/javascript/JContent.register.jsx
@@ -40,7 +40,7 @@ export default function () {
             path: `/sites/${site}`,
             language: language
         }, {
-            requiredPermission: ['pagesAccordionAccess', 'contentFolderAccordionAccess', 'mediaAccordionAccess', 'additionalAccordionAccess', 'formAccordionAccess']
+            requiredSitePermission: [JContentConstants.accordionPermissions.pagesAccordionAccess, JContentConstants.accordionPermissions.contentFolderAccordionAccess, JContentConstants.accordionPermissions.mediaAccordionAccess, JContentConstants.accordionPermissions.additionalAccordionAccess, JContentConstants.accordionPermissions.formAccordionAccess]
         });
 
         if (permissions.loading) {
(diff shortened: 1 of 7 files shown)


gflores-jahia:
Converted to draft for now; need to revisit when story is put back in queue

src/javascript/JContent.register.jsx
```diff
@@ -34,14 +35,40 @@ export default function () {
...
+            requiredSitePermission: [JContentConstants.accordionPermissions.pagesAccordionAccess, JContentConstants.accordionPermissions.contentFolderAccordionAccess, JContentConstants.accordionPermissions.mediaAccordionAccess, JContentConstants.accordionPermissions.additionalAccordionAccess, JContentConstants.accordionPermissions.formAccordionAccess]
+        });
+
+        if (permissions.loading) {
+            return 'Loading...';
+        }
+
+        let defaultMode = JContentConstants.mode.PAGES;
+
+        if (!permissions.node.site.pagesAccordionAccess && !permissions.node.site.contentFolderAccordionAccess && !permissions.node.site.mediaAccordionAccess && !permissions.node.site.formAccordionAccess && !permissions.node.site.additionalAccordionAccess) {
```
gflores-jahia, line 52: Maybe cleaner and more easier to understand if we check for the granted access rather than the negation. I'm thinking of something like:

```
let defaultMode = '';
if (permissions.node.site.pagesAccordionAccess) {
    defaultMode = JContentConstants.mode.PAGES;
} else if (permissions.node.site.contentFolderAccordionAccess) {
    defaultMode = JContentConstants.mode.CONTENT_FOLDERS;
} else if () {
    ....
}
```

gflores-jahia (changes requested):

src/javascript/JContent.register.jsx
XiangYuHe: Make sense to me (My bad making it more complicated then neccessary)

src/javascript/JContent/JContent.constants.js
```diff
@@ -24,6 +26,13 @@ const JContentConstants = {
     },
     localStorageKeys: {
         filesSelectorMode: 'jcontent_files_selector_mode'
+    },
+    accordionPermissions: {
+        pagesAccordionAccess: 'pagesAccordionAccess',
+        contentFolderAccordionAccess: 'contentFolderAccordionAccess',
+        mediaAccordionAccess: 'mediaAccordionAccess',
+        additionalAccordionAccess: 'additionalAccordionAccess',
+        formAccordionAccess: 'formAccordionAccess'
```
gflores-jahia, line 35: maybe we don't need this one since it's been moved to forms?

XiangYuHe: We still need this as on the "jContent button" button click action is still in jContent repository (generic code)
We need to check here for the form access permissions.

gflores-jahia: ok good to know, thanks

gflores-jahia (approved):

PR #407, opened 2021-05-03
BACKLOG-15811: Add-in the drag-and-drop for fileupload.

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15811

## Description

Changes so that the drop zone will display "Nothing to display" if mode is MEDIA (accordion item)

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentLayout/EmptyDropZone/EmptyDropZone.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/EmptyDropZone/EmptyDropZone.jsx
@@ -6,6 +6,8 @@ import {compose} from '~/utils';
 import {withTranslation} from 'react-i18next';
 import JContentConstants from '../../../JContent.constants';
 import {Publish} from '@jahia/moonstone';
+import {useSelector} from 'react-redux';
+import {useNodeChecks} from '@jahia/data-helper';
 
 const styles = theme => ({
     dropZone: {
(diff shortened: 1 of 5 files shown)


src/javascript/JContent/ContentRoute/ContentLayout/EmptyDropZone/EmptyDropZone.jsx
```diff
@@ -37,7 +39,19 @@ const styles = theme => ({
 });
 
 const EmptyDropZone = ({component: Component, t, classes, mode}) => {
-    if (mode === JContentConstants.mode.MEDIA) {
+    const currentState = useSelector(state => ({site: state.site, language: state.language}));
+    const permissions = useNodeChecks({
+        path: `/sites/${currentState.site}`,
+        language: currentState.language
+    }, {
+        requiredSitePermission: ['uploadFilesAction']
```
AKarmanov, line 47: Maybe we could put this in .constants file so that we don't duplicate it in UploadTransformComponent?

src/javascript/JContent/ContentRoute/ContentLayout/EmptyDropZone/EmptyDropZone.jsx
```diff
@@ -46,18 +60,18 @@ const EmptyDropZone = ({component: Component, t, classes, mode}) => {
         );
     }
 
-    if (mode === JContentConstants.mode.PAGES) {
+    if (mode === JContentConstants.mode.CONTENT_FOLDERS) {
```
AKarmanov, line 63: Do we need to have permissions check here as well or no?

XiangYuHe: We do as for CONTENT_FOLDERS we use import instead of upload as text
Will add that permission in my next PR

src/javascript/JContent/ContentRoute/ContentLayout/EmptyDropZone/EmptyDropZone.jsx
XiangYuHe: Sure ya make sense let me consolidate that into 1 variable into constants

AKarmanov (approved):

PR #434, opened 2021-07-19
BACKLOG-16282 Set timeout on preloaded data to make sure it ends up i…

AKarmanov:
…n a different batch and doesn't cause concurrency issues

## JIRA

https://jira.jahia.org/browse/BACKLOG-16282

## Description

Send request in a different batch to avoid concurrency issues


gflores-jahia (commented):
I'm ok with this for now but let's expand more/note down the details on the concurrency issue and what's causing it (either here in description or in the ticket), in case we want to track root cause in the future.

gflores-jahia (approved):

AKarmanov:
> I'm ok with this for now but let's expand more/note down the details on the concurrency issue and what's causing it (either here in description or in the ticket), in case we want to track root cause in the future.

I'll add a comment on the ticket

PR #442, opened 2021-07-28
BACKLOG-16296 Protect against undefined values

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16296

## Description

Prevent undefined exception


gflores-jahia (commented):
Should we add a default blank value (e.g. `r?.displayName || ''`) or undefined is ok?

AKarmanov:
> Should we add a default blank value (e.g. `r?.displayName || ''`) or undefined is ok?

Undefined seems to be ok, we get an empty entry in the table.

gflores-jahia (approved):

PR #448, opened 2021-08-04
BACKLOG-16326 - Reverse order of lint calls to apply the max warning to the js issues

rknj:
The option `--max-warnings` is added by the Orb:
https://github.com/Jahia/jahia-modules-orb/blob/fcc401a2c940d6e9a2eabd73cbf82fb5838482a4/src/commands/lint.yml#L27

And the output is: `yarn lint:js && yarn lint:scss --max-warnings 1`
Meaning the option is applied on the scss check


rknj:
Now we have 54 warnings to fix

gflores-jahia (approved):

PR #466, opened 2021-10-05
BACKLOG-16326: Fixed lint issues

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16326

## Description

Fixed all remaining lint issue.
- I'm still allowing "enabled" prop name instead of "isEnabled", as it would require changes in other modules declaring actions ( like content-editor ). All other boolean props renamed to is*
- For method parameter limit: use an object as parameter
- Dependency array on useEffect / useXx was often invalid and depending on calculated object, I had to add multiple useMemo to avoid the infinite reloads
- reduced complexity by extracting methods

### DIFF
--- a/src/javascript/JContent/actions/publishAction.jsx
+++ b/src/javascript/JContent/actions/publishAction.jsx
@@ -7,7 +7,7 @@ import {useSelector} from 'react-redux';
 import {useTranslation} from 'react-i18next';
 import {setRefetcher, unsetRefetcher} from '../JContent.refetches';
 
-function checkAction(res, node, publishType, allLanguages) {
+function checkAction(res, node, publishType, isAllLanguages) {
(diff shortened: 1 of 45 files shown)


src/javascript/JContent/actions/publishAction.jsx
```diff
@@ -36,13 +36,13 @@ function checkAction(res, node, publishType, allLanguages) {
     return {enabled, isVisible};
 }
 
-function checkActionOnNodes(res, publishType, allLanguages) {
+function checkActionOnNodes(res, publishType, isAllLanguages) {
```
cedmail, line 39: Same type, **isPublishingAllLanguages** will be fitting better I think

cedmail (approved):

PR #467, opened 2021-10-25
BACKLOG-16326 - Fix eslint setup issue and enable the tests

rknj:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16326

## Description

Fix eslint issue by upgrading to latest orb version (the new version is configured to use the LTS version versus a static 14.15x version of node)
I also took the opportunity to enable the Cypress test execution


tdraier (approved):

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 orbs:
   docker: circleci/docker@0.5.20
   jq: circleci/jq@2.2.0
-  jahia-modules-orb: jahia/jahia-modules-orb@0.4.0
+  jahia-modules-orb: jahia/jahia-modules-orb@0.6.2
```
Fgerthoffert: Can you update the orb to use 0.6.4 ?

.circleci/config.yml
```diff
@@ -243,29 +244,29 @@ workflows:
...
+    jobs:
+      - jahia-modules-orb/initialize
+      - jahia-modules-orb/integration-tests:
+          name: Nightly-<< matrix.jahia_image >>
+          context: QA_ENVIRONMENT
+          requires:
+            - jahia-modules-orb/initialize
+          matrix:
+            parameters:
+              jahia_image: ["<< pipeline.parameters.JAHIA_IMAGE >>"]
```
Fgerthoffert, line 264: I'm wondering if we should use the parameter here, versus directly the name:

`jahia_image: ["jahia/jahia-discovery:8", "jahia/jahia-discovery-dev:8-SNAPSHOT"]`

Parameters are useful for providing default values when making API calls (so that you don't need to specify every single parameters)

tests/env.debug.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
Fgerthoffert, line 1: Is there a lot of value in a script that just wait for Jahia and start cypress? 
Is it just to make sure we don't forget to source ?

Fgerthoffert: It's there as well :) , @tdraier is that script really useful?

rknj: this script is not harmful and could ease the local test flow

Fgerthoffert: I'm fine either way

tdraier (approved):

tests/env.debug.sh
tdraier: maybe not very useful but does not harm a lot :-)

PR #468, opened 2021-10-25
TECH-242 : Replace material-ui by moonstone wherever possible

tdraier:
## JIRA

https://jira.jahia.org/browse/TECH-242

## Description

- replaced Typography
- replaced Buttons
- replaced Icons - some icons are still coming from mdi-material-ui or @jahia/icons (material-based)
- Styles : replaced all JSS by SCSS
- Replaced some HOC by hooks in functional components
- Use path alias


ffffffelix (commented):
Good work for this huge PR!
I’m not sure for all spacings conversions and for some button variants. As it’s huge PR it’s difficult to test so I propose to go like that and we will see later if we need to adapt some spacings or variants.

ffffffelix (approved):

PR #508, opened 2022-05-12
QA-14105: Fix DocumentViewer.scss to properly scroll and fit in preview drawer

cedmail:
https://jira.jahia.org/browse/QA-14105

Document viewer is overlapping the preview drawer and hiding data

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,8 +97,7 @@
     "underscore": "1.12.1",
     "xlsx": "^0.17.0",
     "mammoth": "^1.4.13",
-    "node-fetch": "^2.6.1",
-    "minimist": "^1.2.6"
+    "node-fetch": "^2.6.1"
   },
   "devDependencies": {
     "@apollo/react-testing": "^3.1.3",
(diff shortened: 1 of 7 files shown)


gflores-jahia (approved):

package.json
```diff
@@ -97,7 +97,8 @@
     "underscore": "1.12.1",
     "xlsx": "^0.17.0",
     "mammoth": "^1.4.13",
-    "node-fetch": "^2.6.1"
+    "node-fetch": "^2.6.1",
+    "minimist": "^1.2.6"
```
gflores-jahia, line 101: Probably better if we update minimist in yarn.lock instead

gflores-jahia: Edit: I see that yarn.lock is already updated so this is probably not needed

tests/package.json
```diff
@@ -30,5 +30,8 @@
   },
   "dependencies": {
     "@jahia/jahia-reporter": "^1.0.3"
+  },
+  "resolutions": {
+    "minimist" : "^1.2.6"
```
gflores-jahia, line 35: we can probably remove this already since minimist dependencies should already resolve to 1.2.6

gflores-jahia (approved):

PR #517, opened 2022-06-08
BACKLOG-17188: display time in Last Modified Date column

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-17188

Changed format for last modified date column from ll to lll rest of UI is using LLL
https://day.js.org/docs/en/display/format#localized-formats


gflores-jahia (approved):

gflores-jahia:
I'm assuming column width doesn't need to be adjusted since it's longer texts now? (I believe those are fixed)

cedmail:
Tested with english and french seems to fit in same size

rknj:
Wait until end of 8.1.2 code freeze before merging that

PR #519, opened 2022-06-09
BACKLOG-17183 Replace view mode buttons with a dropdown

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-17183

## Description

Replace view mode buttons with a dropdown. Note that original notes in the ticket should not include the last two cases for pickers.


gflores-jahia (approved):

gflores-jahia:
Probably need to wait for 8.1.2 release before merging cc @rknj

PR #527, opened 2022-06-23
Revert Cypress yarn.lock while trying to debug the issue the hanging issue

Fgerthoffert:
See some investigations here: https://github.com/cypress-io/cypress/issues/17627


Fgerthoffert:
jContent does not seem to be hanging anymore, closing the PR for now.

PR #529, opened 2022-06-30
BACKLOG-17222 Expose ContentNavigation, ContentTree and SiteSwitcher

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-17222

## Description

Expose ContentNavigation, ContentTree and SiteSwitcher for use in CE and potentially other modules. Started a branch in CE as well, PR will come later.


src/javascript/JContent/ContentNavigation/ContentNavigation.container.jsx
```diff
@@ -1,47 +1,68 @@
...
+import NavigationHeader from '~/JContent/ContentNavigation/NavigationHeader';
 
-const ContentNavigationContainer = ({mode, siteKey, handleNavigation}) => {
-    let accordionItems = registry.find({type: 'accordionItem', target: 'jcontent'});
-
-    const currentState = useSelector(state => ({site: state.site, language: state.language}));
+const ContentNavigationContainer = ({handleNavigation, selectorObject, accordionItemTarget, accordionItemType, header, sitePermissions}) => {
+    const selector = useMemo(() => createStructuredSelector(selectorObject), [selectorObject]);
+    const dispatch = useDispatch();
+    const {siteKey, language, mode} = useSelector(state => selector(state));
```
tdraier, line 15: I don't understand the usage of reselect here and the format of the selectorObject - why not directly passing the selector function (arg of useSelector) as a prop ? default value would be `state => ({site: state.site, language: state.language, mode: state.jcontent.mode})` , and content editor picker could pass `state => ({site: state.contenteditor.picker.site,lang: state.language, path: state.contenteditor.picker.path,openPaths: state.contenteditor.picker.openPaths})`

The memoizations that reselect brings is not of any value here as we just read some properties into the state, no heavy calculation is involved. It will actually memoize the 3 individual selector (site, language, mode) and the structured one for no benefit.

src/javascript/JContent/ContentNavigation/ContentNavigation.container.jsx
```diff
@@ -1,47 +1,68 @@
...
-    mode: PropTypes.string.isRequired,
-    siteKey: PropTypes.string.isRequired
+    selectorObject: PropTypes.shape({
+        mode: PropTypes.func.isRequired,
+        siteKey: PropTypes.func.isRequired,
+        language: PropTypes.func.isRequired
+    }),
+    accordionItemTarget: PropTypes.string,
+    accordionItemType: PropTypes.string,
+    handleNavigation: PropTypes.func,
```
tdraier, line 50: Maybe rename to `handleNavigationAction`  to understand we are passing a redux action

src/javascript/JContent/ContentNavigation/ContentNavigation.jsx
```diff
@@ -2,12 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import {Accordion, SecondaryNav} from '@jahia/moonstone';
 import {useTranslation} from 'react-i18next';
-import NavigationHeader from './NavigationHeader';
 
-const ContentNavigation = ({accordionItems, mode, siteKey, handleNavigation}) => {
-    const {t} = useTranslation('jcontent');
+const ContentNavigation = ({accordionItems, mode, siteKey, handleNavigation, header}) => {
+    const {t} = useTranslation();
```
tdraier, line 7: you can probably leave 'jcontent' as default namespace, that won't have any impact if a ns is specified in the label

src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/SiteSwitcher.jsx
```diff
@@ -43,48 +40,49 @@ class SiteSwitcher extends React.Component {
...
-        this.props.selectSite(siteNode, newLang);
-    }
+    return newLang ? newLang : siteNode.site.defaultLanguage;
+};
+
+const SiteSwitcher = ({selectorObject, onSelectAction, notificationContext}) => {
+    const {t} = useTranslation();
+    const selector = useMemo(() => createStructuredSelector(selectorObject), [selectorObject]);
+    const dispatch = useDispatch();
+    const {siteKey, currentLang} = useSelector(state => selector(state));
```
tdraier, line 76: Same remark as in ContentNavigation - directly passing the selector function and avoiding reselect will be more clear

src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/SiteSwitcher.jsx
```diff
@@ -119,46 +117,38 @@ class SiteSwitcher extends React.Component {
...
+        currentLang: state => state.language
+    }
 };
 
-export default compose(
-    withTranslation(),
-    connect(mapStateToProps, mapDispatchToProps),
-    withNotifications()
-)(SiteSwitcher);
+export default compose(withNotifications())(SiteSwitcher);
```
tdraier, line 154: Even if not related, maybe it's worth doing it now - if you upgrade to react-material 3.0.5, you can also use useNotifications()  and get rid of all hoc

tdraier (commented):
The code looks much cleaner and the shared components props are quite clear, also nice to get rid of some old class components ..
However I don't think reselect (which is a great lib, btw) brings anything except complexity and a lot of useless memoizations. See inline comments

src/javascript/JContent/ContentNavigation/ContentNavigation.container.jsx
AKarmanov: Not a fan of hook overload myself, I used reselect primarily because they claim to return a memoized object which prevents rerendering on every state change which happens if we simply supply an object or function supplying an object. I don't know how significant the change is in our context but if we are ok with some extra rerendering I can get rid of reselect.

src/javascript/JContent/ContentNavigation/ContentNavigation.container.jsx
```diff
@@ -1,47 +1,68 @@
...
+    const dispatch = useDispatch();
+    const {siteKey, language, mode} = useSelector(state => selector(state));
     const permissions = useNodeChecks({
-        path: `/sites/${currentState.site}`,
-        language: currentState.language
+        path: `/sites/${siteKey}`,
+        language: language
     }, {
-        requiredSitePermission: [JContentConstants.accordionPermissions.pagesAccordionAccess, JContentConstants.accordionPermissions.contentFolderAccordionAccess, JContentConstants.accordionPermissions.mediaAccordionAccess, JContentConstants.accordionPermissions.additionalAccordionAccess, JContentConstants.accordionPermissions.formAccordionAccess]
+        requiredSitePermission: sitePermissions
```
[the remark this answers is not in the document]
AKarmanov: I'll do that.

src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/SiteSwitcher.jsx
AKarmanov: Sure, I'll do the upgrade.

src/javascript/JContent/ContentNavigation/NavigationHeader/SwitchersLayout/SiteSwitcher.jsx
AKarmanov: replied above

src/javascript/JContent/ContentNavigation/ContentNavigation.container.jsx
tdraier: I think I get your point - useSelector will probably trigger a re-render on every redux state change - and the reselect should skip these rerender with the memoization. However this is a problem we have on a lot of useSelector usage (when we return an object), and maybe we should address that more globally in a dedicated ticket. We would need to check the balance between avoiding render and adding memo. 
Also a more simple solution (without memo), when we are just getting {siteKey,language,mode}, would be to pass an equalityFn to useSelector ( from https://react-redux.js.org/api/hooks#equality-comparisons-and-updates , `const selectedData = useSelector(selectorReturningObject, shallowEqual)`)

AKarmanov: Ok, sounds good, then I will remove reselect :)

tdraier (approved):

PR #536, opened 2022-07-14
BACKLOG-17368: Fix left alignment of objects

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-17368

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
+++ b/src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
@@ -1,0 +1,7 @@
+.expansePanel {
+    padding: 24px 24px 32px 0 !important;
+}
+
+.button {
+    margin-left: 0 !important;
+}
--- a/src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
+++ b/src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
@@ -1,7 +1,7 @@
-.expansePanel {
-    padding: 24px 24px 32px 0 !important;
-}
+div.expansePanel {
+    padding: 24px 24px 32px 0;
 
-.button {
-    margin-left: 0 !important;
+    .button {
+        margin-left: 0;
+    }
 }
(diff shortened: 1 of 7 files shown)


src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
```diff
@@ -0,0 +1,7 @@
+.expansePanel {
+    padding: 24px 24px 32px 0 !important;
```
gflores-jahia, line 2: We should probably look at how we can set without using `!important` here. (we can usually override if we use with more specific selectors)

src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
```diff
@@ -0,0 +1,7 @@
...
+}
+
+.button {
+    margin-left: 0 !important;
```
gflores-jahia, line 6: We should probably look at how we can set without using `!important` here. (we can usually override if we use with more specific selectors)

esp. this one as `.button` looks a bit more generic and not sure if that can affect other buttons

gflores-jahia (changes requested):

src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
cedmail: those classes are rewritten/renamed by SCSS compiler so you usually put names that make sense for your local component as there will be no side effect
issue is that moonstone define same properties, material ui also so there is already 2 classes with padding. we need !important to be sure it's this one applying

gflores-jahia: Fixed with https://github.com/Jahia/jcontent/pull/536/commits/12ef5999fdfdeec3027d00dd55e061197788190d

src/javascript/JContent/actions/editImage/ImageEditorDialog/ImageEditorActions/ImageEditorActions.scss
gflores-jahia: Fixed with https://github.com/Jahia/jcontent/pull/536/commits/12ef5999fdfdeec3027d00dd55e061197788190d

gflores-jahia (approved):

PR #566, opened 2022-08-17
BACKLOG-19909 Update canDisplayItem method to use either node or fold…

AKarmanov:
…er, sort ancestors by path descending

## JIRA

https://jira.jahia.org/browse/BACKLOG-19909

## Description

Update canDisplayItem method 
Sort ancestors so that we always get closest one

### DIFF
--- a/src/javascript/JContent/JContent.accordion-items.jsx
+++ b/src/javascript/JContent/JContent.accordion-items.jsx
@@ -18,6 +18,9 @@ import {Sql2SearchQueryHandler} from '~/JContent/ContentRoute/ContentLayout/quer
 
 const filesRegex = /^\/sites\/[^/]+\/files((\/.*)|$)/;
 const contentsRegex = /^\/sites\/[^/]+\/contents((\/.*)|$)/;
+const contentFolderRegex = /^\/sites\/[^/]+\/contents.*/;
+const folderRegex = /^\/sites\/[^/]+\/files.*/;
+const everythingUnderSitesRegex = /^\/sites\/.*/;
 
 export const jContentAccordionItems = registry => {
     const getPath = (site, pathElements, registryItem) => {
@@ -84,7 +87,12 @@ export const jContentAccordionItems = registry => {
                 .filter(n => n.primaryNodeType.name === 'jnt:page');
             return pages[pages.length - 1].path;
         },
-        canDisplayItem: ({selectionNode, folderNode}) => selectionNode ? !filesRegex.test(selectionNode.path) && !contentsRegex.test(selectionNode.path) : /^\/sites\/.*/.test(folderNode.path),
+        canDisplayItem: ({selectionNode, folderNode}) =>
+            selectionNode ? !filesRegex.test(selectionNode.path) &&
+                !contentsRegex.test(selectionNode.path) :
+                everythingUnderSitesRegex.test(folderNode.path) &&
+                !folderRegex.test(folderNode.path) &&
+                !contentFolderRegex.test(folderNode.path),
         getViewTypeForItem: node => node.primaryNodeType.name === 'jnt:page' ? 'pages' : 'content',
         requiredSitePermission: JContentConstants.accordionPermissions.pagesAccordionAccess,
         config: {
@@ -106,7 +114,7 @@ export const jContentAccordionItems = registry => {
         icon: <FolderSpecial/>,
         label: 'jcontent:label.contentManager.navigation.contentFolders',
         defaultPath: siteKey => '/sites/' + siteKey + '/contents',
-        canDisplayItem: ({selectionNode, folderNode}) => selectionNode ? contentsRegex.test(selectionNode.path) : /^\/sites\/[^/]+\/contents.*/.test(folderNode.path),
+        canDisplayItem: ({selectionNode, folderNode}) => selectionNode ? contentsRegex.test(selectionNode.path) : contentFolderRegex.test(folderNode.path),
         requiredSitePermission: JContentConstants.accordionPermissions.contentFolderAccordionAccess,
         config: {
             rootPath: '/contents',
@@ -124,7 +132,7 @@ export const jContentAccordionItems = registry => {
         icon: <Collections/>,
         label: 'jcontent:label.contentManager.navigation.media',
         defaultPath: siteKey => '/sites/' + siteKey + '/files',
-        canDisplayItem: ({selectionNode, folderNode}) => selectionNode ? filesRegex.test(selectionNode.path) : /^\/sites\/[^/]+\/files.*/.test(folderNode.node),
+        canDisplayItem: ({selectionNode, folderNode}) => selectionNode ? filesRegex.test(selectionNode.path) : folderRegex.test(folderNode.node),
         requiredSitePermission: JContentConstants.accordionPermissions.mediaAccordionAccess,
         config: {
             rootPath: '/files',
--- a/src/javascript/JContent/JContent.accordion-items.jsx
+++ b/src/javascript/JContent/JContent.accordion-items.jsx
@@ -18,7 +18,6 @@ import {Sql2SearchQueryHandler} from '~/JContent/ContentRoute/ContentLayout/quer
 
 const filesRegex = /^\/sites\/[^/]+\/files((\/.*)|$)/;
 const contentsRegex = /^\/sites\/[^/]+\/contents((\/.*)|$)/;
-const folderRegex = /^\/sites\/.*/;
 
 export const jContentAccordionItems = registry => {
     const getPath = (site, pathElements, registryItem) => {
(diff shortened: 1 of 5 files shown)


src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -51,7 +64,7 @@ export const jContentAccordionItems = registry => {
         getPath,
         getUrlPathPart,
         getPathForItem: node => {
-            return node.ancestors[node.ancestors.length - 1].path;
+            return node.ancestors.sort(compareNodesByPathDesc)[node.ancestors.length - 1].path;
```
tdraier, line 67: why do you have to sort ancestors ? they are always anyway sorted by path length (maybe in the opposite order ?)

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -80,10 +93,12 @@ export const jContentAccordionItems = registry => {
...
         defaultPath: siteKey => '/sites/' + siteKey,
         getPathForItem: node => {
-            const pages = node.ancestors.filter(n => n.primaryNodeType.name === 'jnt:page');
+            const pages = node.ancestors
+                .filter(n => n.primaryNodeType.name === 'jnt:page')
+                .sort(compareNodesByPathDesc);
             return pages[pages.length - 1].path;
         },
-        canDisplayItem: node => !filesRegex.test(node.path) && !contentsRegex.test(node.path),
+        canDisplayItem: (selectionNode, folderNode) => selectionNode ? !filesRegex.test(selectionNode.path) && !contentsRegex.test(selectionNode.path) : folderRegex.test(folderNode.path),
```
tdraier, line 101: if only one of the 2 params is expected, maybe use `({selectionNode, folderNode})` instead ?

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -105,7 +120,7 @@ export const jContentAccordionItems = registry => {
         icon: <FolderSpecial/>,
         label: 'jcontent:label.contentManager.navigation.contentFolders',
         defaultPath: siteKey => '/sites/' + siteKey + '/contents',
-        canDisplayItem: node => contentsRegex.test(node.path),
+        canDisplayItem: (selectionNode, folderNode) => selectionNode ? contentsRegex.test(selectionNode.path) : folderRegex.test(folderNode.path),
```
tdraier, line 123: not sure about the generic folderRegex - media can display /sites/xx/files , /sites/xx/files/xxx, but not /sites/xx/contents or  /sites/xx/contents/xxx - and only pages can display /sites/xx . here folderRegex matches all of them ?

src/javascript/JContent/JContent.accordion-items.jsx
AKarmanov: yes, right! they are actually sorted

src/javascript/JContent/JContent.accordion-items.jsx
AKarmanov: That's true, I'll have to do a dedicated one

AKarmanov:
I actually have to update all the calls now that signature has changed

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -80,10 +80,11 @@ export const jContentAccordionItems = registry => {
         label: 'jcontent:label.contentManager.navigation.pages',
         defaultPath: siteKey => '/sites/' + siteKey,
         getPathForItem: node => {
-            const pages = node.ancestors.filter(n => n.primaryNodeType.name === 'jnt:page');
+            const pages = node.ancestors
+                .filter(n => n.primaryNodeType.name === 'jnt:page');
             return pages[pages.length - 1].path;
         },
-        canDisplayItem: node => !filesRegex.test(node.path) && !contentsRegex.test(node.path),
+        canDisplayItem: ({selectionNode, folderNode}) => selectionNode ? !filesRegex.test(selectionNode.path) && !contentsRegex.test(selectionNode.path) : /^\/sites\/.*/.test(folderNode.path),
```
tdraier, line 87: still in the case of folderNode , this one should not match the ones that are matched by contentFolders and media - like /sites/xx/files (so not match the 2 regex from content-folder and media ?)

AKarmanov:
Added a check for jnt:folder and jnt:contentFolder on primary node type, this should be sufficient to know which case we are in (I think it should that is).

src/javascript/JContent/JContent.utils.js
```diff
@@ -168,3 +168,16 @@ export const arrayValue = value => {
...
+
+export const getCanDisplayItemParams = node => {
+    const folders = ['jnt:contentFolder', 'jnt:folder'];
+    const params = {};
+
+    if (folders.includes(node.primaryNodeType.name)) {
+        params.folderNode = node;
+    } else {
+        params.selectionNode = node;
+    }
```
tdraier, line 180: the switch between folderNode / selectionNode does not depend on the type, but rather on the context on which the method is being called. A "locate" action will always try to locate a selection node (expandTree is trying to find for the item a possible parent in the available accordions), and the "open in jcontent" will always try to find a folderNode (just looking for an available accordion for the passed node)

src/javascript/JContent/actions/expandTree.jsx
```diff
@@ -28,7 +29,8 @@ const GetAncestorsQuery = gql`
 export const expandTree = (path, client) => {
     return client.query({query: GetAncestorsQuery, variables: {path}}).then(res => {
         let node = res.data.jcr.nodeByPath;
-        const acc = registry.find({type: 'accordionItem', target: 'jcontent'}).find(acc => acc.canDisplayItem(node));
+        const params = getCanDisplayItemParams(node);
```
tdraier, line 32: so in that case params is always `= {selectionNode:node}`
```suggestion
        const params = {selectionNode: node};
```

src/javascript/JContent/actions/openInJcontentAction.jsx
```diff
@@ -31,7 +32,8 @@ export const OpenInJContentActionComponent = ({path, render: Render, loading: Lo
                 isVisible={res.checksResult}
                 enabled={res.checksResult}
                 onClick={() => {
-                    const acc = registry.find({type: 'accordionItem', target: 'jcontent'}).find(acc => acc.canDisplayItem(res.node));
+                    const params = getCanDisplayItemParams(res.node);
```
tdraier, line 35: and here `={folderNode:res.node}`
```suggestion
                    const params = {folderNode: res.node};
```

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -18,6 +18,9 @@ import {Sql2SearchQueryHandler} from '~/JContent/ContentRoute/ContentLayout/quer
 
 const filesRegex = /^\/sites\/[^/]+\/files((\/.*)|$)/;
 const contentsRegex = /^\/sites\/[^/]+\/contents((\/.*)|$)/;
+const contentFolderRegex = /^\/sites\/[^/]+\/contents.*/;
+const folderRegex = /^\/sites\/[^/]+\/files.*/;
+const everythingUnderSitesRegex = /^\/sites\/.*/;
```
tdraier, line 23: The 2 first regex are actually matching /files and /contents , exactly like the contentFolderRegex and folderRegex (but the 2 latter also matches /filesxx and /contentsxx)

```suggestion
const filesRegex = /^\/sites\/[^/]+\/files\/.*/;
const contentsRegex = /^\/sites\/[^/]+\/contents\/.*/;
const contentFolderRegex = /^\/sites\/[^/]+\/contents((\/.*)|$)/;
const folderRegex = /^\/sites\/[^/]+\/files((\/.*)|$)/;
const everythingUnderSitesRegex = /^\/sites\/.*/;
```

src/javascript/JContent/JContent.accordion-items.jsx
```diff
@@ -80,10 +83,16 @@ export const jContentAccordionItems = registry => {
...
         getPathForItem: node => {
-            const pages = node.ancestors.filter(n => n.primaryNodeType.name === 'jnt:page');
+            const pages = node.ancestors
+                .filter(n => n.primaryNodeType.name === 'jnt:page');
             return pages[pages.length - 1].path;
         },
-        canDisplayItem: node => !filesRegex.test(node.path) && !contentsRegex.test(node.path),
+        canDisplayItem: ({selectionNode, folderNode}) =>
+            selectionNode ? !filesRegex.test(selectionNode.path) &&
+                !contentsRegex.test(selectionNode.path) :
```
tdraier, line 92: maybe also check everythingUnderSitesRegex for the selectionNode
```suggestion
            selectionNode ? everythingUnderSitesRegex.test(selectionNode.path) && 
                !filesRegex.test(selectionNode.path) &&
                !contentsRegex.test(selectionNode.path) :
```

src/javascript/JContent/actions/expandTree.jsx
AKarmanov: This and regex modification brings me back to the issue of locating "files" folder because it is now trying to match selection node with ".../files/.*" while selection node is actually a folder with path ending at ".../files". We have the same issue with ".../contents" folder. That's why I added that type check thinking that it was actually dependent on node type.

src/javascript/JContent/JContent.utils.js
AKarmanov: See my comment for one issue that still have with that locate functionality.

tdraier (approved):

PR #567, opened 2022-08-17
BACKLOG-17381: centralize structured view and result parsing in query…

tdraier:
…Handlers

## JIRA

https://jira.jahia.org/browse/BACKLOG-17381

## Description

Changed the queryHandler so that it directly returns a "result" object with formatted and structured data. Move the logic of structuring from table.container to queryHandler and useLayoutQuery.
Simplified the BaseQueryHandler.getQueryParams so that it takes the same params as the others.
Makes structureData available in BaseQueryHandler to be used in other (including content-editor) handlers.

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentLayout/ContentLayout.container.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/ContentLayout.container.jsx
@@ -14,7 +14,6 @@ import {cmRemoveSelection, cmSwitchSelection} from './contentSelection.redux';
 import {cmSetPreviewSelection} from '~/JContent/preview.redux';
 import ContentLayout from './ContentLayout';
 import {refetchTypes, setRefetcher, unsetRefetcher} from '~/JContent/JContent.refetches';
-import {isInSearchMode, structureData} from '../ContentLayout/ContentLayout.utils';
 import {Loader} from '@jahia/moonstone';
 import {useLayoutQuery} from '~/JContent/ContentRoute/ContentLayout/useLayoutQuery';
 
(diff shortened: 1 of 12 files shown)


src/javascript/JContent/ContentRoute/ContentLayout/ContentLayout.container.jsx
```diff
@@ -159,17 +158,6 @@ export const ContentLayoutContainer = () => {
...
-        if (data && data.jcr && data.jcr.nodeByPath) {
-            // When new results have been loaded, use them for rendering.
-            let nodeTypeName = data.jcr.nodeByPath.primaryNodeType.name;
-            let subTypes = ['jnt:page', 'jnt:contentFolder', 'jnt:virtualsite'];
-            let isSub = !subTypes.includes(nodeTypeName);
-            // Sub is not the same as params.sub; refresh and sync up path param state
-            if (isSub !== (params.sub === true)) { // Params.sub needs to be boolean type; else falsy
-                setPath(path, {sub: isSub});
-            }
-        }
```
gflores-jahia, line 171: just to confirm, but I didn't see this refactored anywhere. is this not needed anymore (specifically `setPath`)?

tdraier: yes, good catch :-) actually this code was supposed to detect when the `sub` parameter was incorrectly set, but I did not manage to find a real use case. The sub parameter is correctly passed on links, false on pages/folder and true on content. However, it's keeping switching from `sub` to false then back to true when browsing a content (so 2 setPath operations which breaks the navigation history). 
as it's using data.jcr.nodeByPath which is not available anymore, i prefered to get rid of it ..

gflores-jahia (approved):

PR #651, opened 2022-11-23
Modified label for live button

sdusolle:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20114

## Description

Modified the live button label into "Live" as requested in the Figma.


tdraier:
It seems the label is missing in fr and de ?

tdraier (approved):

PR #659, opened 2022-11-28
BACKLOG-20359: Rename fragment

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20359


gflores-jahia (commented):
Does the naming here also need to change? https://github.com/Jahia/jcontent/blob/master/src/javascript/JContent/ContentTree/ContentTree.jsx#L49

I see a reference to `PrimaryNodeTypeName` there as well

tdraier:
Good catch, yes, I think it's the same name and should be renamed too

gflores-jahia (approved):

PR #662, opened 2022-11-30
BACKLOG-20414: Improve dragndrop design

ffffffelix:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20414

## Description
- Design the target element for drop/insertion in Page Composer. Use the same design as other drag-and-drop styles.
- Render the jContent table as a target for dropping instead of a `div`  below the table
- Use the same typography weight in the empty area for the default state and the "drop target" state

### DIFF
--- a/src/javascript/JContent/PageComposerRoute/EditFrame/Box.scss
+++ b/src/javascript/JContent/PageComposerRoute/EditFrame/Box.scss
@@ -1,3 +1,6 @@
+$color-accent_light: rgba(99, 209, 255, 0.6);
+$color-accent: rgb(0, 160, 227);
+
 .root {
     position: absolute;
     z-index: 9999;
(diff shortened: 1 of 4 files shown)


src/javascript/JContent/PageComposerRoute/EditFrame/Box.scss
```diff
@@ -1,3 +1,6 @@
+$color-accent_light: rgba(99, 209, 255, 0.6);
+$color-accent: rgb(0, 160, 227);
```
ffffffelix, line 2: I don't know why the moonstone CSS properties don't work here.
It seems the others moonstone variables used in this file are associated with unused CSS.

tdraier: The moonstone variables are not copied into the iframe dom, so the css which are used there cannot use them

tdraier (approved):

PR #828, opened 2023-05-19
BACKLOG-20990: Restore paste action on jnt:page and jnt:navMenuText

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20990

## Description

Restore paste action on jnt:page and jnt:navMenuText


gflores-jahia:
Would this cause regression on https://jira.jahia.org/browse/BACKLOG-20680?

cedmail:
> Would this cause regression on https://jira.jahia.org/browse/BACKLOG-20680?

Should not I could not figure why we were hiding actions on jnt:page

gflores-jahia (approved):

PR #838, opened 2023-05-26
QA-14511: harmonize publication queries to get updated cache

tdraier:
## JIRA

https://jira.jahia.org/browse/QA-14511


gflores-jahia (approved):

gflores-jahia (commented):
Test failing looks to be same one as one master, so not sure if related or not.

PR #868, opened 2023-06-26
BACKLOG-21119: Introduce back the maxNameSize check for file uploads.

jahia-carp:
https://jira.jahia.org/browse/BACKLOG-21119


tdraier:
Shouldn't we update the default value to a more decent size, at least 64 chars ?

cedmail:
> Shouldn't we update the default value to a more decent size, at least 64 chars ?

tdraier (approved):

PR #888, opened 2023-07-03
TECH-1324 Add actionBars test

Janin-Michel-Mathias:
## JIRA

https://jira.jahia.org/browse/TECH-1324

## Description
Add actionBars test


dgriffon (changes requested):
Can you use ID or class names to validate the element to be present? Labels can change where IDs should less.

dgriffon (approved):

PR #891, opened 2023-07-05
Use scripts from cypress

tdraier:


cedmail (changes requested):
We need jahia-ui-root latest when testing against jahia released version on nightly

cedmail (dismissed):

PR #912, opened 2023-08-01
BACKLOG-20402: Increase integration test timeouts; Fix missing test module build

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20402

At least until we get a handle of test runtime, then we can adjust


tests/assets/provisioning.yml
```diff
@@ -37,3 +35,7 @@
   site: "digitall"
 
 - executeScript: "script-01-create-categories.groovy"
+
+- uninstallBundle: 
+    - "org.jahia.modules/content-editor/4.6.0.SNAPSHOT"
+    - "org.jahia.modules/jcontent/2.12.0.SNAPSHOT"
```
tdraier, line 41: you can probably omit the version here

tdraier (dismissed):

tests/assets/provisioning.yml
gflores-jahia: thanks. remove it from CE. also I don't think we need to uninstall jcontent since it gets updated anyways.

gflores-jahia:
Also adjusting max warnings for static analysis for now until we stabilize tests, then we can refactor and readjust.

tdraier (dismissed):

PR #916, opened 2023-08-02
Test updated actions

rknj:
## JIRA

https://jira.jahia.org/browse/TECH-1384


gflores-jahia:
Uses `main` branch :)

PR #921, opened 2023-08-03
QA-14909: Area restrictions defined at template level are found on parent

cedmail:
## JIRA

https://jira.jahia.org/browse/QA-14909

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


tdraier:
I don't understand this - even without the fix, if I have a sublist under another list, and a constraint set on the main list, this constraint is "cascaded" under the sublist. This should not be the case (it's not the case in gwt page composer)

tdraier (dismissed):

tdraier (approved):

PR #976, opened 2023-08-22
BACKLOG-21467: Make JContent 3.0 compatible 8.2 only, nightly uses matrix. Changes test-reports name

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-2146

## Description

Make JContent 3.0 compatible 8.2 only, nightly uses matrix. Changes test-reports name

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   integration-tests-standalone:
-    name: Integration Tests (Standalone)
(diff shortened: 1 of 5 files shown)


gflores-jahia (commented):
should we also update parent version to 8.2? https://github.com/Jahia/jcontent/blob/master/pom.xml#L32

cedmail:
> should we also update parent version to 8.2? https://github.com/Jahia/jcontent/blob/master/pom.xml#L32

I did already

gflores-jahia (dismissed):

PR #987, opened 2023-08-28
TECH-1328 add dateTimePicker and fix datePicker

Janin-Michel-Mathias:
## JIRA

https://jira.jahia.org/browse/TECH-1328

## Description
add dateTimePicker and fix datePicker

### DIFF
--- a/tests/cypress/e2e/pickers/dateTime.cy.ts
+++ b/tests/cypress/e2e/pickers/dateTime.cy.ts
@@ -1,0 +1,82 @@
+import {createSite, createUser, grantRoles, addNode, getNodeByPath, deleteUser, deleteSite, deleteNodeProperty} from '@jahia/cypress';
+import {JContent} from '../../page-object/jcontent';
+import {DateTimePicker} from '../../page-object/dateTimePicker';
+import {ContentEditor} from '../../page-object';
+
+const visitDateTimePicker = () => {
+    cy.get('@dateTimePicker').then(uuid => {
+        cy.visit(`/jahia/jcontent/severalLanguages/fr/content-folders/contents#(contentEditor:!((formKey:modal_0,isFullscreen:!t,lang:fr,mode:edit,uuid:'${uuid}')))`);
+    });
+};
+
+const deleteAndCheck = () => {
+    const dateTimePicker = new DateTimePicker();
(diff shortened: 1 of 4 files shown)


tests/cypress/page-object/dateTimePicker.ts
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        const stringHour = hourPartsInt[0] < 10 ? `0${hourPartsInt[0]}` : hourPartsInt[0];
+        const stringMinutes = hourPartsInt[1] < 10 ? `0${hourPartsInt[1]}` : hourPartsInt[1];
+
+        return `${stringHour}${separator}${stringMinutes}`;
+    }
+
+    public getTimePicker(): Cypress.Chainable {
+        return cy.get('input[id="qant:pickers_datetimepicker"]');
```
jsinovassin, line 40: `qant:pickers_datetimepicker` should be set as parameter, otherwise we will not be able to use this function for other node types

jsinovassin: As the id is used in several function of this class, you can set the value of the id as a global property that you set when you call the constructor of this object

tests/cypress/page-object/datePicker.ts
```diff
@@ -1,5 +1,5 @@
 export class DatePicker {
-    private getTodayDate(separator: string): string {
+    protected getTodayDate(separator: string): string {
```
jsinovassin, line 2: I know that it's already merged but it could be nice to have qant:pickers_datepicker as a parameter instead of an harcoded value in getDatePicker()

tests/cypress/e2e/pickers/dateTime.cy.ts
```diff
@@ -0,0 +1,82 @@
...
+        dateTimePicker.checkDate('');
+        dateTimePicker.typeNowTime();
+        cy.get('body').click();
+        dateTimePicker.checkNowTime();
+        contentEditor.save();
+        dateTimePicker.checkNowTime();
+        deleteAndCheck();
+    });
+
+    afterEach('Check Value is kept after saving and clean picker', () => {
```
jsinovassin, line 74: Maybe we can change the text here to just `Logout` as it's the only thing which is done in the afterEach

jsinovassin (changes requested):

jsinovassin (approved):

PR #988, opened 2023-08-29
QA-14898 : add language to be used in session language in graphql API

jsinovassin:
https://jira.jahia.org/browse/QA-14898


cedmail:
Tests about content creation are all broken, as this PR depends on https://github.com/Jahia/graphql-core/pull/336

tdraier:
It's not very clear to me which language we're supposed to pass here. With jcontent - and graphql api in general - a content can be edited in multiple languages at the same time. It seems we need to randomly pass a language in this "sessionLanguage" (basically, the last one being edited). The change in graphql makes the api very unclear, as nobody can understand what would be the impact of this sessionLanguage. 
In my opinion the issue is more in the rule itself - if it needs to test that "a" language exists, it has to check on all possible languages - otherwise it has to specify the language.

PR #1012, opened 2023-09-12
BACKLOG-21541 Add simple test

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21541

## Description

Add test


tests/cypress/e2e/jcontent/breadcrumbsPageBuilder.cy.ts
```diff
@@ -0,0 +1,31 @@
+import {JContent, JContentPageBuilder} from '../../page-object';
+
+describe('Breadcrumbs in page builder', () => {
```
cedmail, line 3: Can we specify it's boxes breadcrumbs in the label as there is already tests for breadcrumbs

cedmail (changes requested):

cedmail (approved):

PR #1022, opened 2023-09-19
BACKLOG-21552: Clear selection on copy/cut action

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21552

## Description

Enable paste action right after a cut/copy by clearing the selection


gflores-jahia:
Re-doing implementation

PR #1056, opened 2023-10-26
BACKLOG-20186 Use basic toolbar for ckeditor

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-20186

## Description

Use basic toolbar for ckeditor

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/graphql/api/forms/GqlEditorForms.java
+++ b/src/main/java/org/jahia/modules/contenteditor/graphql/api/forms/GqlEditorForms.java
@@ -168,9 +168,7 @@ public class GqlEditorForms {
         String toolbar = "Light";
         try {
             JCRNodeWrapper node = getSession().getNode(nodePath);
-            if (node.hasPermission("view-full-wysiwyg-editor")) {
-                toolbar = "Full";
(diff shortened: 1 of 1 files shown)


src/main/java/org/jahia/modules/contenteditor/graphql/api/forms/GqlEditorForms.java
```diff
@@ -168,9 +168,7 @@ public String ckeditorToolbar(@GraphQLName("nodePath") @GraphQLDescription("node
         String toolbar = "Light";
         try {
             JCRNodeWrapper node = getSession().getNode(nodePath);
-            if (node.hasPermission("view-full-wysiwyg-editor")) {
-                toolbar = "Full";
-            } else if (node.hasPermission("view-basic-wysiwyg-editor")) {
+            if (node.hasPermission("view-full-wysiwyg-editor") || node.hasPermission("view-basic-wysiwyg-editor")) {
                 toolbar = "Basic";
             }
```
tdraier, line 173: Are we ok that also root and other power users will loose access to the full editor ? 
My understanding was that we just wanted to change the permissions for the editor and translators, from view-full-wysiwyg-editor to view-basic-wysiwyg-editor.

PR #1102, opened 2023-11-23
BACKLOG-21852: Ensure we used up-to-date information for pickers

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21852


gflores-jahia (approved):
should we take out the console.debug?

or it's something we can start removing through webpack: https://medium.com/@shaangontia/remove-console-statements-webpack-5-38455e4b471a

cedmail:
console.debug do not show up by default in the console and we don't do anything costly so I think we should keep them

PR #1119, opened 2023-12-08
BACKLOG-21965: Extension point for external pickers

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21965

## Description
 
Based on https://github.com/Jahia/content-editor/tree/BACKLOG-21926-mvp 

- New extension point "externalPickerConfiguration" allows to provide alternative ui for specific pickers
- PickerDialog adds a selector on top when external picker configurations are detected
- Moved the effective Jahia picker dialog code to "JahiaPicker"
- PickerDialog delegates to either JahiaPicker or external picker component
- ContentPickerApi allows for value more formats - path, uuid, or "url", or array of them
- PickerDialog now always takes an array of objects ([{url,path,uuid}]) for "initialSelectedItem"
- PickerDialog can return an array of nodes of array of {url} objects.
- RichText now allows to have URLs inserted and edited (instead of nodes identified by path), and passed to picker


tests/cypress/e2e/pickers/externalpicker.cy.ts
```diff
@@ -0,0 +1,38 @@
...
+    it('should display content reference picker', () => {
+        const pickerField = jcontent
+            .createContent('Pickers')
+            .getPickerField('qant:pickers_filepicker');
+        const picker = pickerField.open();
+        getComponentByRole(Dropdown, 'picker-selector').select('Custom picker');
+        cy.contains('Test picker');
+        cy.get('button[data-sel-role="custom-selector"]').click();
+        pickerField.assertValue('user.jpg');
+    });
```
jkevan, line 37: We could add a test on a rich text ensuring that RichText is correctly handling full URL when selecting an image from the test DAM ?

I know we can't test references as it would imply to implement an DAM EDP along the picker, but rich text seem's testable, what do you think ?

jkevan (dismissed):

jkevan (approved):

PR #1129, opened 2023-12-18
BACKLOG-21345: Add categoryManagerRightClick tests

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21345

## Description

- Refactor category manager to a separate page object
- Merge scenarios described in story from selenium to cypress with existing tests

### DIFF
--- a/tests/cypress/e2e/jcontent/categoryManager.cy.ts
+++ b/tests/cypress/e2e/jcontent/categoryManager.cy.ts
@@ -1,44 +1,69 @@
 import {CategoryManager} from '../../page-object';
-const accordionItemName = 'category';
+import {deleteNode} from "@jahia/cypress";
+
 describe('Category Manager', () => {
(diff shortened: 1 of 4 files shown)


cedmail (dismissed):

sdusolle (dismissed):

gflores-jahia:
Found an issue regarding selecting a tree item element with a given depth. 

It looks like the actual style text is dependent on the Chrome version - version 107 (the one we use for cypress image tests) has the attribute `style="--tree-depth:2;"` while in the latest Chrome (currently v120) it shows up as `style="--tree-depth: 2;` (with space). Which is why it was passing locally but not when running the PR.

Added a workaround to work for both scenarios, but I will be creating a PR in moonstone to add a proper data-depth attribute to our tree item elements

sdusolle (approved):

PR #1134, opened 2023-12-22
BACKLOG-22113: Do not show header if mouse is over a new element

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22113

## Description

Do not show header if mouse is over a new element and previous element was pinned


gflores-jahia (dismissed):
Weird...I was expecting the onmouseout would be setting that header to false, whether pinned or not, but it somehow never gets that event. I guess safer to do it on the mouseover.

PR #1156, opened 2024-01-15
CI/CD: Fixed incorrect path in the HTML tests report

Fgerthoffert:
This is a part of a change done in CE: https://github.com/Jahia/content-editor/commit/bec3ef0225c61b9832f73aa1366b1648461d9ebe#diff-c7273ca77f452520adba2dfb738cd146b83110e7faa7b861318e89b2b56b5d56

As-is, the videos can't load because the path is incorrect
Screenshot 2024-01-15 at 15 10 28


rknj (dismissed):

Fgerthoffert:
Actually, fixing is slightly more complicated than I first thought.

For some reason, Cypress is not saving the file in a subfolder of the same name than the subfolder containing the spec. The reporter is following the correct pattern, but cannot find the video.

PR #1161, opened 2024-01-29
BACKLOG-21980 Introduce async lifecycle hooks 

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21980

## Description

Introduce async lifecycle hooks

### DIFF
--- a/src/javascript/ContentEditor/ContentEditor/useOnBeforeContextHooks.js
+++ b/src/javascript/ContentEditor/ContentEditor/useOnBeforeContextHooks.js
@@ -1,0 +1,22 @@
+import {useState, useEffect, useRef} from 'react';
+import {registry} from '@jahia/ui-extender';
+import {useApolloClient} from '@apollo/client';
+
+export const useOnBeforeContextHooks = editorContext => {
+    const client = useApolloClient();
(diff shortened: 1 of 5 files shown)


gflores-jahia (approved):

gflores-jahia:
Had discussion with @AKarmanov regarding adding new entrypoint for resolving richtext editor vs using https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/SelectorTypes/resolveSelectorType.js#L7:

> This would only work if I could hardcode richtext5 mixin related info in jcontent pass it down to that selector and create a resolution function on RichText selector which would use all that info and conditionally return selector from richtext 5 module.
> either do it backend (via some form of form/field interceptor and replace selector) or front end before we even load the editor to avoid flickering and issues with async nature of the check. I had backend working but then it seemed too much and there is no other use for that interceptor.
> another issue is that our resolution is sync so any call we do would do something after initial loading, changing it to async is quite a task as it is used in a few places and is kind of critical

Not to mention registry is lacking lifecycle phases for us to do clean up when ckeditor5 module is uninstalled. So we'll go with this option for now.

PR #1162, opened 2024-01-30
BACKLOG-22280: Add test for displaying resource bundle values on disabled module

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22280


gflores-jahia:
failing tests already in nightly

cedmail (approved):

PR #1175, opened 2024-03-05
TECH-1634: Use new subscription entry point for publication job. Remove registration of PushEventHandler

cedmail:
…

## JIRA

https://jira.jahia.org/browse/TECH-1634

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -6,6 +6,10 @@ schema {
     subscription: Subscription
 }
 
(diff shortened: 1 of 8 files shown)


gflores-jahia (commented):
Code looks ok to me, but seems to have failure on this test: 
```Test the save publish buttons flow Check save publish buttons flow```

I don't see it on nightly, so maybe something related to the change?

PR #1176, opened 2024-03-08
BACKLOG-22170: Remove deprecated GWT panels from Advanced tabs

cedmail:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22170


gflores-jahia (commented):
I don't see DeprecatedOption component used anymore. We can probably clean that out as well.

gflores-jahia (approved):

PR #1181, opened 2024-03-12
TECH-1657: Replace jahia-discovery with jahia-ee

rknj:
## JIRA

https://jira.jahia.org/browse/TECH-1657

## Description

- Updated the main provisioning script to deploy all the required modules
- Updated the manual-run to use jahia-ee instead of jahia-discovery
- Added the required parameter for dependency track


cedmail (dismissed):

gflores-jahia (commented):
Compared to https://github.com/Jahia/jahia-pack/blob/master/additional-modules/pom.xml, there's a few missing modules but I'm also not sure if they are required for the tests or not:

- article/3.0.0
- contact/3.1.0-SN
- faq/3.0.0
- googleAnalytics/3.0.0
- socialsharing/3.0.0

Also mismatched versions to latest 8.2-SN (it could be needed for 8.2 compatibility):

- demo-components/2.5-SN
- demo-core/2.6-SN
- demo-templates/3.4
- event/4-SN
- rating/3.3-SN

and 2 modules where additional-modules seems to be behind (I guess this is ok):

- bootstrap3-core/4.2.1
- bootstrap3-components/4.2.1

gflores-jahia (approved):

PR #1200, opened 2024-04-02
BACKLOG:22519: fixed translations in CE

sdusolle:
## Description

I fixed the translations for the "publish" button in CE in French and German. It was appearing in English.


sdusolle:
NOTE : All integration tests are failing because we removed digitall and we did not yet update the provisioning

rknj (approved):

PR #1209, opened 2024-04-16
BACKLOG-22134: Update version and add missing Nexus env var

rknj:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22134

### DIFF
--- a/.github/workflows/on-code-change.yml
+++ b/.github/workflows/on-code-change.yml
@@ -47,6 +47,8 @@ jobs:
     name: Build Module
     needs: update-signature
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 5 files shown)


gflores-jahia (commented):
Probably need a signature update as well

rknj:
the signature has not been triggered by JahiaReleaseTool so I think we're good

gflores-jahia:
It was updated from recent change here https://github.com/Jahia/jcontent/pull/1208

rknj:
I did it manually

gflores-jahia (dismissed):
might be 401 due to staging repos

gflores-jahia (approved):

PR #1220, opened 2024-05-22
TECH-740: Addressed code smells

r3dm1ke:
## JIRA

https://jira.jahia.org/browse/TECH-740

## Description

This PR fixes some common code smells reported by sonar.

## Tests

- [ ] Integration Tests

## Checklist

- [x] Performance
- [x] Migration
- [x] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanel.refetches.js
+++ b/src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanel.refetches.js
@@ -1,7 +1,7 @@
 const editPanelRefetches = {};
 
-let triggerRefetch = key => {
-    let refetch = editPanelRefetches[key];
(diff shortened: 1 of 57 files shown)


gflores-jahia (commented):
It looks like some of the changes are 1) causing issues with eslint rules and 2) might actually introduced some bugs (due to test case failures). 

I don't know if the eslint issues are because of conflicts with the sonarcube ones and if so, we need to figure out how to consolidate these two, or settle on using one or the other.

gflores-jahia:
Just adding that it seems we also have some issue with provisioning which could be causing tests failures as well.

We should update this branch when PR https://github.com/Jahia/jcontent/pull/1222 has been merged to master and re-run

gflores-jahia (approved):

PR #1221, opened 2024-05-22
QA-14892: Use rootPath for pages accordion if nothing else available

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-14892

## Description

Backport https://github.com/Jahia/jcontent/pull/1143

### DIFF
--- a/src/javascript/JContent/JContent.accordion-items.jsx
+++ b/src/javascript/JContent/JContent.accordion-items.jsx
@@ -57,7 +57,9 @@ export const jContentAccordionItems = registry => {
             return path;
         },
         getPathForItem(node) {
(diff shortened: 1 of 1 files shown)


gflores-jahia:
I'm a bit confused on this backport. We do not have this change in master: https://github.com/Jahia/jcontent/blob/master/src/javascript/JContent/JContent.accordion-items.jsx#L65

I'm assuming it's backporting this one https://github.com/Jahia/jcontent/pull/1143? but looks to be in a different spot in the code. 

Should the change be done here instead? https://github.com/Jahia/jcontent/blob/2_x/src/javascript/JContent/JContent.accordion-items.jsx#L110

gflores-jahia (approved):

PR #1245, opened 2024-07-17
QA-15051 Adapt i18n fields

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-15051

## Description

Adapt i18n fields

Removed unittest in order to avoid unnecessary maintenance, we test this component in cypress with copyFromOneLanguageToCurrentLanguage.cy.ts


gflores-jahia (approved):

src/javascript/ContentEditor/actions/contenteditor/copyLanguage/copyLanguage.utils.js
```diff
@@ -17,27 +18,24 @@ export function getFullLanguageName(languages, id) {
...
         .filter(field => field.i18n)
-        .map(field => {
-            return {
-                name: field.name,
-                multiple: field.multiple
-            };
-        });
+        .map(field => ({...field, propertyName: field.name}));
+
+    const adaptedValues = i18nFields.map(field => getFieldValues(field, jcr?.result)).reduce((acc, current) => ({...acc, ...current}), {});
```
jkevan, line 29: Overal the method: getI18nFieldAndValues seem's a bit clunky, even before your modifications:
- we could adapt field on demand in the return statement to avoid adapting fields that could be filtered few lines later.
- I count 3 iterations on i18nFields, seem's a lot for what we need to do here. (2x `i18nFields.map`, 1x `i18nFields.find`) I think the algorythms could be simplified here and be more efficient and also more easy to read.

Consider:
In the return statement, when doing: `jcr.result.properties.filter` getting the associated field at the same time and adapt on the fly the value.

PR #1252, opened 2024-07-30
BACKLOG-22629 Add tests module release workflow

Janin-Michel-Mathias:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22629

## Description
Add tests module release workflow

### DIFF
--- a/.github/workflows/release-tests-module.yml
+++ b/.github/workflows/release-tests-module.yml
@@ -1,0 +1,23 @@
+name: Publish tests module to Registry
+on:
+    workflow_dispatch:
+
+
+jobs:
+  npm_publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+      - run: cd tests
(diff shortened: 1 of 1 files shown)


tests/cypress/page-object/pageBuilder.ts
```diff
@@ -0,0 +1,14 @@
+import {BasePage} from '@jahia/cypress';
+
+export class PageBuilder extends BasePage {
```
jkevan, line 3: Not sure to understand why a new class ? Why not reusing and returning a ContentEditor instance ?
Like it's the case for the editPage: https://github.com/Jahia/jcontent/blob/master/tests/cypress/page-object/pageComposer.ts#L133-L139
```
 editPage(title: string): ContentEditor {
        cy.iframe('#page-composer-frame', this.iFrameOptions).within(() => {
            cy.get('#JahiaGxtPagesTab').contains(title).rightclick({force: true});
            cy.get('.pagesContextMenuAnthracite').contains('Edit').click({force: true});
        });
        return new ContentEditor();
    }
```

jkevan (changes requested):

jkevan (approved):

PR #1261, opened 2024-08-05
BACKLOG-22873: Bump moonstone to 2.7.3

rknj:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22873


rknj:
Resolution of react-file-viewer/**/ajv has been automatically removed. Note that it has been added with https://github.com/Jahia/jcontent/commit/0e5f7d1443d022ab863467c55fc3faad11ed24ac (TECH-1166) but without any explicit reason

PR #1273, opened 2024-08-15
QA-15251: Fix URL parsing in rich text ckeditor

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/QA-15251

## Description

- Include protocol when parsing URL links in ckeditor
- Add checks when parsing URL links in ckeditor for file and content prefix paths
- Add jest tests and modify cypress test to include breakage check with existing URL

### DIFF
--- a/tests/cypress/e2e/pickers/richtext.cy.ts
+++ b/tests/cypress/e2e/pickers/richtext.cy.ts
@@ -37,6 +37,8 @@ describe('Picker tests - richtext', () => {
         richText.type('Hello');
 
         const linkModal = richText.openLinkModal();
+        // Test to make sure parsing of existing URLs do not break picker
+        linkModal.getURLField().type('https://www.interregemr.eu/projets/marhetak-fr');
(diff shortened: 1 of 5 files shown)


tests/cypress/e2e/pickers/richtext.cy.ts
```diff
@@ -37,6 +37,8 @@ describe('Picker tests - richtext', () => {
         richText.type('Hello');
 
         const linkModal = richText.openLinkModal();
+        // Test to make sure parsing of existing URLs do not break picker
+        linkModal.getURLField().type('https://www.interregemr.eu/projets/marhetak-fr');
```
gflores-jahia, line 41: It used to be the parsing of this specific url leads to the string `/projets/markehtak-pr`. Because of the starting slash, it is then treated as a jcr path where query throws an error and causes issues. Test to make sure this is not happening anymore.

AKarmanov (approved):

PR #1289, opened 2024-09-10
QA-15174 Preview now honors the j:view property

r3dm1ke:
## JIRA

https://jira.jahia.org/browse/QA-15174

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentLayout/PreviewDrawer/Preview/PreviewComponent/PreviewComponent.container.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/PreviewDrawer/Preview/PreviewComponent/PreviewComponent.container.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {useContentPreview} from '@jahia/data-helper';
+import {useContentPreview, useNodeInfo} from '@jahia/data-helper';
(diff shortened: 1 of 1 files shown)


gflores-jahia:
while this works, I think part of the implementation notes is that we wanted to make this change on the graphql layer so that we avoid querying stuff that's already available in the graphql. Specifically this additional note on the story:

> We should update the GraphQL endpoint renderedContent so that it uses the view set on the node by default instead of using the default view.
> If a specific view is passed in the GraphQL call, of course this view should be used.
> 
> => in terms of priority when calling renderedContent:
> 1 - use the view passed in the graphql call, if there is one
> 2 - use the view set on the node, if there is one
> 3 - use the default view

We also might want to rewrite the PR that's already merged in master

gflores-jahia:
After discussion from daily, because we want to limit the backport dependencies and create delays for 8.1, we are going with the UI approach at least for jcontent 2x, as oppose to using the fix available in graphql (although we are still backporting this part in graphql 2x for 8.1 still).

gflores-jahia (approved):

PR #1297, opened 2024-09-18
QA-15011 Update deletion usage info, return empty for mimetype if not found

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-15011

## Description

Update deletion usage info, return empty for mimetype if not found

### DIFF
--- a/src/javascript/JContent/actions/deleteActions/Delete/Delete.jsx
+++ b/src/javascript/JContent/actions/deleteActions/Delete/Delete.jsx
@@ -84,11 +84,11 @@ const DeleteContent = ({data, onClose, isLoading, isMutationLoading, dialogType,
                 <Loader size="big" style={{width: '100%'}}/> :
                 <DialogContent>
                     <DialogContentText className={styles.content} dangerouslySetInnerHTML={{__html: label}}/>
-                    {hasUsages && count === 1 &&
+                    {!usagesOverflow && hasUsages && count === 1 &&
                         <DialogContentText>{t('jcontent:label.contentManager.deleteAction.hasUsages.single')}</DialogContentText>}
-                    {hasUsages && (count > 1 || count === 0) &&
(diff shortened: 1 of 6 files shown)


src/javascript/JContent/actions/deleteActions/Delete/Delete.jsx
```diff
@@ -86,7 +86,7 @@ const DeleteContent = ({data, onClose, isLoading, isMutationLoading, dialogType,
                     <DialogContentText className={styles.content} dangerouslySetInnerHTML={{__html: label}}/>
                     {hasUsages && count === 1 &&
                         <DialogContentText>{t('jcontent:label.contentManager.deleteAction.hasUsages.single')}</DialogContentText>}
-                    {hasUsages && count > 1 &&
+                    {hasUsages && (count > 1 || count === 0) &&
```
gflores-jahia, line 89: trying to understand how we can have usages but with 0 count. is this because of permissions? 

`count` conditional also doesn't look like it's necessary here and can be removed, but also needs to be replaced by a check if theres usagesOverflow (otherwise case below won't trigger)

gflores-jahia: ok understood with the permissions.

The count conditional is only checking if it's a non-negative number which will always be the case (the singular case is already taken care of above), base on the count calculation. Even in the case where count ends up with an invalid value (e.g. count=undefined or NaN) maybe it's safer to display this as a fail-safe message as long as hasUsages is true.

also ignore the usagesOverflow comment, I misread read the conditional as `hasUsages && usagesOverflow`

gflores-jahia: after discussion, we've simplified the conditional but still need it here as we can end up with two messages when count == 1

gflores-jahia (approved):

PR #1315, opened 2024-10-04
QA-15313 Do not check for jnt:navMenuText when deciding selectability

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-15313

## Description

 Do not check for jnt:navMenuText when deciding selectability

### DIFF
--- a/src/javascript/JContent/ContentTree/ContentTree.gql-fragments.js
+++ b/src/javascript/JContent/ContentTree/ContentTree.gql-fragments.js
@@ -26,7 +26,7 @@ const PickerItemsFragment = {
             language: 'String!'
         },
         gql: gql`fragment IsTreeSelectable on JCRNode {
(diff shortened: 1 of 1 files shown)


gflores-jahia:
This looks it might cause a regression. It seems it was added ([commit](https://github.com/Jahia/jcontent/pull/649)) to prevent error when clicking on a nav menu item in jcontent nav (see comment [here](https://jira.jahia.org/browse/BACKLOG-20115?focusedId=142765&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-142765)). We might need some way to maybe make this a picker-only setting.

gflores-jahia:
Discussed with @AKarmanov and no regression seen in this case, so change should be good.

gflores-jahia (approved):

PR #1317, opened 2024-10-07
BACKLOG-23160 Preload and map areas

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-23160

## Description

Preload and map areas, note that this also fixes "nothing shown in structured view on page creation" issue.

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentRoute.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentRoute.jsx
@@ -28,6 +28,8 @@ export const ContentRoute = () => {
     const res = useNodeInfo({path}, {getIsNodeTypes: nodeTypes});
     const {FLAT, STRUCTURED, PAGE_BUILDER} = JContentConstants.tableView.viewMode;
     const accordionItem = registry.get('accordionItem', mode);
+    const isPageBuilderView = viewMode === PAGE_BUILDER;
+    const canShowEditFrame = res?.node && nodeTypes.some(nt => res.node[nt]);
 
     useEffect(() => {
         if (accordionItem.tableConfig?.availableModes?.indexOf?.(viewMode) === -1) {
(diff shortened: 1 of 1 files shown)


src/javascript/JContent/ContentRoute/ContentRoute.jsx
```diff
@@ -32,6 +35,38 @@ export const ContentRoute = () => {
         }
     }, [dispatch, mode, viewMode, FLAT, accordionItem]);
 
+    // Captured area information is used to block delete/move/copy/cut actions on areas
+    useEffect(() => {
+        if (path && language) {
```
gflores-jahia, line 40: Should we maybe add a check around here to see if a node is renderable on page builder or not, as not all nodes are not always renderable? For example visiting pages/home/area-main/highlights where it is not renderable in page builder.

FYI there is already a check happening around line 50 for this so maybe we just need to move this whole block below it.

src/javascript/JContent/ContentRoute/ContentRoute.jsx
```diff
@@ -32,6 +35,38 @@ export const ContentRoute = () => {
...
+            loadPageAndCaptureJahiaAreas(path, language, template);
+        }
+    }, [path, language, template]);
+
+    const loadPageAndCaptureJahiaAreas = (path, language, template) => {
+        const renderMode = 'editframe';
+        const encodedPath = path.replace(/[^/]/g, encodeURIComponent) + (template === '' ? '' : `.${template}`);
+        const url = `${window.contextJsParameters.contextPath}/cms/${renderMode}/default/${language}${encodedPath}.html?redirect=false`;
```
gflores-jahia, line 48: Should we also maybe use renderUrl graphql API here vs building url manually? Not sure if we're going to encounter issues here similar to this : https://jira.jahia.org/browse/QA-15228

gflores-jahia: yeah I think we're probably referring to the same story, I was looking at the PR you recently did here https://github.com/Jahia/jcontent/pull/1250/files?w=1 and was using renderUrl, but not sure if that was done specifically for fixing live urls. Maybe your way is more appropriate if it's for edit mode only, I'm not sure, but I just wanted to make sure the part around servername/urls are also all taken care of (when using diff sites etc).

gflores-jahia (approved):

PR #1321, opened 2024-10-17
QA-14926 : add ChoiceTree selector type in jContent

dgriffon:
This PR adds a new selector type to jContent : `ChoiceTree`
This selector is the same as the current category tree but with configurable options:
- rootPath : allows to define the root node of the tree
- types:  node types displayed in the tree. 

As the selector type is not an allowed one by Jahia core (https://github.com/Jahia/jahia/blob/ff486af706d236e07587d4ee49d5c6cbb0afd751/core/src/main/java/org/jahia/services/content/nodetypes/SelectorType.java#L73-L85) you need a json override to use it:
```
    {
      "name": "multipleChoiceTreeWithTypes",
      "selectorType": "ChoiceTree",
      "selectorOptionsMap": {
        "rootPath": "/sites/choiceTreeSelectorSite/contents",
        "types": "cent:testChoiceTreeContent,cent:testChoiceTreeContentRoot"
      }
```
For example:
https://github.com/Jahia/jcontent/blob/d053fb38b8b83f10772f35b576fa583855f5c091/tests/jahia-module/jcontent-test-module/src/main/resources/META-INF/jahia-content-editor-forms/fieldsets/cent_testChoicetree.json

The definition must be multiple to display the selector type 

Covered by cypress tests.


src/javascript/ContentEditor/SelectorTypes/ChoiceTree/ChoiceTree.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+    const tree = choiceTreeAdapter({
+        nodes: data.jcr.result.descendants.nodes,
+        parent: data.jcr.result,
+        selectedValues: value,
+        locale: editorContext.lang
+    });
+
+    const singleValue = field.multiple ? undefined : value;
+    const multipleValue = field.multiple ? (value || []) : undefined;
```
jkevan, line 64: I don't see why we have single vs multiple logic here, couple of line upper there is already:
```
if (!field.multiple) {
        console.warn('ChoiceTree selector is not supporting single value');
        return <></>;
}
```

jkevan (changes requested):

src/javascript/ContentEditor/SelectorTypes/Category/Category.jsx
```diff
@@ -1,77 +1,20 @@
...
-            size="medium"
-            value={singleValue}
-            values={multipleValue}
-            isDisabled={field.readOnly}
-            onClear={handleClear}
-            onChange={handleChange}
-            onBlur={onBlur}
-        />
-    );
+    return <ChoiceTree field={field} value={value} id={id} editorContext={editorContext} onChange={onChange} onBlur={onBlur} />;
```
jkevan, line 17: Are you sure we can use it ? Looking at the Category.jsx it's seem it was able to works on single valued field.
But it's seem the new ChoiceTree is now only able to work with multiple valued field.

src/javascript/ContentEditor/SelectorTypes/ChoiceTree/ChoiceTree.jsx
dgriffon: single logic has been preserved

src/javascript/ContentEditor/SelectorTypes/Category/Category.jsx
dgriffon: single logic has been preserved

PR #1325, opened 2024-10-30
BACKLOG-23210 Flush page and site cache

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-23210

## Description

Flush page and site cache. Address some minor button, styling issues

### DIFF
--- a/src/javascript/JContent/ContentRoute/MainActionBar/MainActionBar.jsx
+++ b/src/javascript/JContent/ContentRoute/MainActionBar/MainActionBar.jsx
@@ -35,8 +35,8 @@ export const MainActionBar = () => {
                 <>
                     { !isFolder && (
                         <ButtonGroup size="big" variant="default" color="accent" className={styles.item}>
-                            <DisplayAction isMediumLabel actionKey="openInPreview" path={path} isDisabled={isDisabled} render={ButtonRendererShortLabel} buttonProps={{variant: 'outlined', size: 'big', className: styles.item}}/>
-                            <DisplayAction menuUseElementAnchor actionKey="openInPreviewMenu" path={path} isDisabled={isDisabled} render={ButtonRendererNoLabel} buttonProps={{variant: 'outlined', size: 'big', color: 'accent', icon: <ChevronDown/>}}/>
+                            <DisplayAction isMediumLabel actionKey="openInPreview" path={path} isDisabled={isDisabled} render={ButtonRendererShortLabel} buttonProps={{variant: 'ghost', size: 'big', className: styles.item}}/>
+                            <DisplayAction menuUseElementAnchor actionKey="openInPreviewMenu" path={path} isDisabled={isDisabled} render={ButtonRendererNoLabel} buttonProps={{variant: 'ghost', size: 'big', color: 'accent', icon: <ChevronDown/>}}/>
                         </ButtonGroup>)}
                     <DisplayAction actionKey="openInLive" path={path} isDisabled={isDisabled} render={ButtonRenderer} buttonProps={{variant: 'outlined', size: 'big', color: 'accent', className: styles.item}}/>
                 </>
--- a/src/javascript/JContent/JContent.actions.jsx
+++ b/src/javascript/JContent/JContent.actions.jsx
@@ -30,8 +30,7 @@ import {
     Undelete,
     Unlock,
     Upload,
-    Visibility,
-    SwapHoriz
+    Visibility
 } from '@jahia/moonstone';
 import {
     ClearAllLocksActionComponent,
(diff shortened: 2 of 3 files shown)


src/main/java/org/jahia/modules/contenteditor/graphql/api/GqlJContentMutations.java
```diff
@@ -0,0 +1,85 @@
...
+ * The root class for the jContent mutation API
+ */
+@GraphQLDescription("jContent API")
+public class GqlJContentMutations {
+
+
+    @GraphQLField
+    @GraphQLName("flushPageCache")
+    @GraphQLDescription("Flushes cache for a page, checks permission and node type")
+    public Boolean flushPageCache(@GraphQLName("pagePath") @GraphQLDescription("Page path") String pagePath) {
```
gflores-jahia, line 45: enforce non null

src/main/java/org/jahia/modules/contenteditor/graphql/api/GqlJContentMutations.java
```diff
@@ -0,0 +1,85 @@
...
+            throw new DataFetchingException(e);
+        }
+
+        return false;
+    }
+
+    @GraphQLField
+    @GraphQLName("flushSiteCache")
+    @GraphQLDescription("Flushes cache for a site, will resolve site node if the path supplied is not a site, checks permission")
+    public Boolean flushSiteCache(@GraphQLName("sitePath") @GraphQLDescription("Site path") String sitePath) {
```
gflores-jahia, line 65: Enforce non null

src/main/java/org/jahia/modules/contenteditor/graphql/api/GqlJContentMutations.java
```diff
@@ -0,0 +1,85 @@
...
+        JCRNodeWrapper node = null;
+
+        try {
+            node = JCRSessionFactory.getInstance().getCurrentUserSession().getNode(sitePath);
+
+            if (node.hasPermission("adminCache")) {
```
gflores-jahia, line 71: Check for node exist

src/main/java/org/jahia/modules/contenteditor/graphql/api/GqlJContentMutations.java
```diff
@@ -0,0 +1,85 @@
...
+    @GraphQLField
+    @GraphQLName("flushPageCache")
+    @GraphQLDescription("Flushes cache for a page, checks permission and node type")
+    public Boolean flushPageCache(@GraphQLName("pagePath") @GraphQLDescription("Page path") String pagePath) {
+        JCRNodeWrapper node = null;
+
+        try {
+            node = JCRSessionFactory.getInstance().getCurrentUserSession().getNode(pagePath);
+
+            if (node.hasPermission("adminCache") && node.isNodeType("jnt:page")) {
```
gflores-jahia, line 51: check for node exist

gflores-jahia, line 51: Do we think maybe we should throw a `GqlAccessDeniedException` instead for the permission check?
If not maybe at least add reasons of why it can get rejected/return false in the description.

src/main/java/org/jahia/modules/contenteditor/graphql/api/GqlJContentMutations.java
gflores-jahia, line 71: Do we think maybe we should throw a `GqlAccessDeniedException` instead for the permission check?
If not maybe at least add reasons of why it can get rejected/return false in the description.

src/main/java/org/jahia/modules/contenteditor/graphql/api/GqlJContentMutations.java
```diff
@@ -0,0 +1,85 @@
...
+    @GraphQLDescription("Flushes cache for a site, will resolve site node if the path supplied is not a site, checks permission")
+    public Boolean flushSiteCache(@GraphQLName("sitePath") @GraphQLDescription("Site path") String sitePath) {
+        JCRNodeWrapper node = null;
+
+        try {
+            node = JCRSessionFactory.getInstance().getCurrentUserSession().getNode(sitePath);
+
+            if (node.hasPermission("adminCache")) {
+                if (!node.isNodeType("jnt:virtualsite")) {
+                    node = JCRContentUtils.getParentOfType(node, "jnt:virtualsite");
```
gflores-jahia, line 73: check for null ?

gflores-jahia (changes requested):
It looks like we're mixing fixes for the edit/live comparison with the flush cache implementation. Could we create a separate PRs for the edit/live and flush cache? This way, it's easier to trace the commits/PR and accurate links with the stories.

gflores-jahia (commented):
New changes looks good.

Sorry, I know it's a pain but I would insist on creating a separate PR for the live/edit action changes. I know it can sometimes happen by accident, but we should try to not mix up our stories with code changes as it makes the history of code changes confusing, we lose track of the ticket source that could have additional discussions, regarding the changes and harder to trace the reason/context on why we made those changes (checking for regression scenarios, etc).

It doesn't have to be 3 separate PRs but we can group them together on a single one as it's all dealing with a single topic.

gflores-jahia (approved):

PR #1336, opened 2024-11-12
BACKLOG-23170 Prevent cut/delete/drag of categories

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-23170

## Description

 Prevent cut/delete/drag of categories


gflores-jahia (commented):
I don't know if this needs clarification, but I believe the story is only asking to prevent the deletion/moving actions for the 'Categories' root folder specifically (as it's system generated and causes issues when deleted), and not generically for all `jnt:category` nodetypes (which I think is what this code change is doing)

image

gflores-jahia:
Also, I'm not sure how the categories root folder is created, but we can maybe make use of the new blockUiAction mixin on creation and block that generically in the categories app

PR #837, opened 2023-05-26
QA-14710: Implement file name checks (max size, invalid chars) during file upload

jkevan:

PR #1338, opened 2024-11-14
QA-15249 Using maxNameSize for renaming validation

r3dm1ke:
## JIRA

https://jira.jahia.org/browse/QA-15249

## Description

Backport from v3. Using the maxNameSize for input length validation when renaming.

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/actions/createFolderAction/CreateFolderDialog/CreateFolderDialog.jsx
+++ b/src/javascript/JContent/actions/createFolderAction/CreateFolderDialog/CreateFolderDialog.jsx
@@ -79,7 +79,7 @@ export const CreateFolderDialog = ({path, contentType, onExit}) => {
                 <TextField
                     fullWidth
                     autoFocus
-                    inputProps={{maxLength: 32}}
+                    inputProps={{maxLength: contextJsParameters.config.maxNameSize}}
                     error={Boolean(errMsg)}
                     value={name}
                     id="folder-name"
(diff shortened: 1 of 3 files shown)


gflores-jahia (commented):
Has this variable been defined somewhere? I'm looking at the change in master and I see that it was being set [here](https://github.com/Jahia/jcontent/pull/837/files?w=1#diff-efb2d7c2f294bf152a48972faf168fe9385bf883b3df83711286f5ae6197b689R6) but I'm not seeing that in 2_x branch.

Also I would recommend changing that as well [here](https://github.com/Jahia/jcontent/blob/2_x/src/javascript/JContent/actions/createFolderAction/CreateFolderDialog/CreateFolderDialog.jsx#L82) as it was also part of that change https://github.com/Jahia/jcontent/pull/606/files?w=1

r3dm1ke:
> Has this variable been defined somewhere? I'm looking at the change in master and I see that it was being set [here](https://github.com/Jahia/jcontent/pull/837/files?w=1#diff-efb2d7c2f294bf152a48972faf168fe9385bf883b3df83711286f5ae6197b689R6) but I'm not seeing that in 2_x branch.
> 
> Also I would recommend changing that as well [here](https://github.com/Jahia/jcontent/blob/2_x/src/javascript/JContent/actions/createFolderAction/CreateFolderDialog/CreateFolderDialog.jsx#L82) as it was also part of that change https://github.com/Jahia/jcontent/pull/606/files?w=1

Good catch about new folder, thank you; and I added the line to the config now;

gflores-jahia (dismissed):

AKarmanov (approved):

gflores-jahia (approved):

PR #1344, opened 2024-11-20
DMF-6336: Allow to add additional parameters in edit frame url

Janin-Michel-Mathias:
## JIRA

https://jira.jahia.org/browse/DMF-6336

## Description
Add states to make the switch between variants possible in page builder

## Tests

- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/src/javascript/JContent/EditFrame/EditFrame.jsx
+++ b/src/javascript/JContent/EditFrame/EditFrame.jsx
@@ -218,7 +218,6 @@ export const EditFrame = ({isDeviceView}) => {
             const framePath = mainModule?.getAttribute('path');
             const locale = mainModule?.getAttribute('locale');
             if (path === framePath && locale === language && previousDevice.current === deviceParam && currentUrlParams === previousUrlParams) {
-                console.log(1);
                 // Clone all styles with doubled classname prefix
                 const head = currentDocument.querySelector('head');
                 iframe.current.ownerDocument.querySelectorAll('style[styleloader],style[data-jss]').forEach(s => {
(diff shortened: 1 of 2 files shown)


src/javascript/JContent/EditFrame/EditFrame.jsx
```diff
@@ -208,13 +210,15 @@ export const EditFrame = ({isDeviceView}) => {
...
+        const url = `${window.contextJsParameters.contextPath}/cms/${renderMode}/default/${language}${encodedPath}.html?redirect=false${deviceParam}${currentUrlParams}`;
+
         if (currentDocument) {
             const mainModule = currentDocument.querySelector('[jahiatype=mainmodule]');
             console.debug('Loading', url, 'in iframe', mainModule?.getAttribute('path'), path, language, deviceParam, previousDevice.current, deviceParam, template);
             const framePath = mainModule?.getAttribute('path');
             const locale = mainModule?.getAttribute('locale');
-            if (path === framePath && locale === language && previousDevice.current === deviceParam) {
+            if (path === framePath && locale === language && previousDevice.current === deviceParam && currentUrlParams === previousUrlParams) {
+                console.log(1);
```
gflores-jahia, line 221: ok to remove?

gflores-jahia (commented):
looks ok to me, but wondering if we should stick with usage of references `useRef` rather than states as all other parts of the iframe url is currently being tracked, at least just to keep it consistent but I'm not quite clear on the usage of it on the `PageHeaderContainer` component

gflores-jahia:
Discussed with @jsinovassin and in this case we are deviating from the `useRef` pattern (e.g. device) because there is a need to refresh the entire document (as a variant) hence the usage of React states instead. Ok with this implementation.

gflores-jahia (dismissed):

gflores-jahia (approved):

PR #1373, opened 2024-12-12
BACKLOG-23452: Provide a way to override date format in date time picker

dgriffon:
Provide a way to override date format in date time selector type using a configuration name forceDateFormat (DD/MM/YYYY, MM/DD/YYYY are only valid format)

https://jira.jahia.org/browse/BACKLOG-23452

(Same as https://github.com/Jahia/content-editor/pull/1778 in content editor)

### DIFF
--- a/src/main/resources/META-INF/configurations/org.jahia.modules.jcontent.cfg
+++ b/src/main/resources/META-INF/configurations/org.jahia.modules.jcontent.cfg
@@ -7,3 +7,7 @@ hideLegacyPageComposer=true
 
 # Allow to define the number of "New Content" buttons when an editing zone as content restrictions activated
 createChildrenDirectButtons.limit = 5
+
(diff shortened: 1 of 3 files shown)


gflores-jahia (approved):

src/main/resources/META-INF/configurations/org.jahia.modules.jcontent.cfg
```diff
@@ -1,4 +1,3 @@
-# default configuration
```
gflores-jahia, line 1: Hi, sorry didn't notice the first time, but it looks like we're potentially overriding existing (customer) changes/customizations with this one. Do we want to do this? or maybe just have this config as optional (enable only if we're overriding).

gflores-jahia (approved):

PR #1389, opened 2025-01-08
BACKLOG-23452: Trim the date format when configured

baptistegrimaud:
https://jira.jahia.org/browse/BACKLOG-23452

## Description

When a date format is configured in _Jcontent_ (`jcontent.forceDateFormat` property), trim its value ~~and convert it to uppercase,~~ to loose restrictions on this property (as the value **must** be `'MM/DD/YYYY'` or  `'DD/MM/YYYY'`).

## Tests

- [x] Unit Tests (Most changes _should_ have unit tests)

### DIFF
--- a/src/javascript/ContentEditor/SelectorTypes/DateTimePicker/DateTimePicker.jsx
+++ b/src/javascript/ContentEditor/SelectorTypes/DateTimePicker/DateTimePicker.jsx
@@ -18,7 +18,7 @@ function getDateFormat(editorContext) {
     const allowedOverridesDateFormat = ['MM/DD/YYYY', 'DD/MM/YYYY'];
 
     // Read date format from config
-    const forceDateFormat = window.contextJsParameters?.config?.jcontent?.forceDateFormat;
+    const forceDateFormat = window.contextJsParameters?.config?.jcontent?.forceDateFormat?.trim().toUpperCase();
     if (forceDateFormat && !allowedOverridesDateFormat.includes(forceDateFormat)) {
         console.warn(`forceDateFormat as been set to an invalid value (${forceDateFormat}). Please use one of the following values: ${allowedOverridesDateFormat.join(', ')}`);
     } else if (forceDateFormat) {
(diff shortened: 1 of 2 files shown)


jkevan (commented):
Good point about using trim(). However, we need to be careful with the handling of uppercase vs. lowercase letters, particularly in date formatting.

**Case Sensitivity in Date Formatting**
In most software libraries, case matters in date formatting because uppercase and lowercase letters typically represent different date or time components. For instance:

`DD/MM/YYYY` and `dd/MM/yyyy` are almost never equivalent:
`DD` often refers to the "day of the year."
`dd` refers to the "day of the month."

**Library-Specific Behavior**
We need to confirm how the date library we are using handles these cases. For example, as per the [Day.js documentation](https://day.js.org/docs/en/display/format):

`DD` → The day of the month, 2-digits (e.g., `01-31`).
`dd` → The short name of the day of the week (e.g., `Su, Mo`).

**Consideration for Configuration or Transformation**
If our configuration deviates from the default behavior of the library (e.g., via custom transformations), it’s worth considering how this could affect user expectations. Users might expect the behavior to align exactly with the library's documentation and could be confused if transformations lead to unexpected results.

**Conclusion**
I’m not entirely sure what the best approach is here, we could also argue that our config is handled by us and we do what we want, and that users should be aware of the underlying library, but still we allow for date format config, so it's a bit strange to transform the date format afterward.

baptistegrimaud:
> Good point about using trim(). However, we need to be careful with the handling of uppercase vs. lowercase letters, particularly in date formatting.
> 
> **Case Sensitivity in Date Formatting** In most software libraries, case matters in date formatting because uppercase and lowercase letters typically represent different date or time components. For instance:
> 
> `DD/MM/YYYY` and `dd/MM/yyyy` are almost never equivalent: `DD` often refers to the "day of the year." `dd` refers to the "day of the month."
> 
> **Library-Specific Behavior** We need to confirm how the date library we are using handles these cases. For example, as per the [Day.js documentation](https://day.js.org/docs/en/display/format):
> 
> `DD` → The day of the month, 2-digits (e.g., `01-31`). `dd` → The short name of the day of the week (e.g., `Su, Mo`).
> 
> **Consideration for Configuration or Transformation** If our configuration deviates from the default behavior of the library (e.g., via custom transformations), it’s worth considering how this could affect user expectations. Users might expect the behavior to align exactly with the library's documentation and could be confused if transformations lead to unexpected results.
> 
> **Conclusion** I’m not entirely sure what the best approach is here, we could also argue that our config is handled by us and we do what we want, and that users should be aware of the underlying library, but still we allow for date format config, so it's a bit strange to transform the date format afterward.

You're right, converting it to uppercase may create confusion for users configuring lowercase date format.
It's safer to simply reject/show a warning if the pattern does not match the supported patterns.
Thanks for your review and your feedback!

jkevan (approved):

PR #1412, opened 2025-01-23
Add German translation for published

sdusolle:
## Description

We are missing a German translation for the "published" status in jContent.

### DIFF
--- a/src/main/resources/javascript/locales/de.json
+++ b/src/main/resources/javascript/locales/de.json
@@ -260,6 +260,8 @@
         "publishMediumPublished": "Veröffentlicht - {{language}}",
         "publishShort": "Veröffentlichen",
         "publishShortPublished": "Veröffentlicht",
+        "publishPublishedMedium": "Veröffentlicht - {{language}}",
+        "publishPublishedShort": "Veröffentlicht",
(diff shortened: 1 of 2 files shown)


src/main/resources/javascript/locales/de.json
```diff
@@ -260,6 +260,8 @@
         "publishMediumPublished": "Veröffentlicht - {{language}}",
         "publishShort": "Veröffentlichen",
         "publishShortPublished": "Veröffentlicht",
+        "publishPublishedMedium": "Veröffentlicht - {{language}}",
+        "publishPublishedShort": "Veröffentlicht",
```
gflores-jahia, line 264: Are these needed? Looks to be duplicates of `publishMediumPublished` and `publishShortPublished` which should be what were using atm. (These were switched from this PR)

If it's breaking, I think we need to open a ticket to switch to the latter.

If anything, these keys also do not exist in en.json but it does in fr.json, so I think we should remove the fr ones.

sdusolle: I removed them from fr and de

gflores-jahia (approved):

PR #1418, opened 2025-01-29
QABACKLOG-1661: Upgraded cypress base image

Fgerthoffert:
[jira.jahia.org/browse/QABACKLOG-1661](https://jira.jahia.org/browse/QABACKLOG-1661)


Fgerthoffert:
Most test suites are passing, the failure is very unlikely related to this PR. 
I'd recommend merging, considering that otherwise everything will fail.

PR #1593, opened 2025-02-14
build: partial revert for syncing test module version on release

gflores-jahia:
## Description

Partial revert from https://github.com/Jahia/jcontent/pull/1425 as it seems to be causing test modules to be deployed now.


sdusolle (approved):

gflores-jahia:
No need to revert. Release issues are not related to this change.

PR #1640, opened 2025-03-11
feat: Page builder content statuses

### LINKED ISSUE #1623: Content statuses (used to be overlays)
The goal of this story is to offer the same functional coverage than what exists in Page Composer (statuses overlays)

**As an** editor

**I want to** have the ability to see particular statuses on the content items present in a page without having to select each one of them one by one.

**In order to** secure and fasten my work

### Acceptance criteria

**Given** I'm in Page Builder (visual mode) 

**When** I ask to show a type of status

**Then** each item  that match the condition is bordered in orange + an icon

That status display remains activated until I manually decide so, even if I navigate and change to another page, or change mode (list/structured).
 and comes back to visual mode.
This setting is at least stored in my session, and it remains activated even between two work sessions.

### Basics

1) When an item has language visible condition, display with grey border and grey overlay + 'Not visible now' badge when displayed in the language where it should not be visible.
2) When a status is marked as "**Work in progress**" it's by default bordered in orange + badge (wrench icon/ label). This border is permanent (until the WIP is removed)
  - It should show only for the given languages if specified
3) When a content item is untranslated (same conditions as in page composer) it is bordered in orange with a block (it's by default it cannot be disabled) 
Use case 1: Something is rendered from the view - display badge "No translation - {{language}}"
Use case 2: Nothing is rendered from the view - display badge "No translation - {{language}}" and with a message "Nothing to display"
If possible in both cases we should display as a block
4) When a content item is marked for deletion, we don't display other statuses
5) When an item is dragging, we hide all status/overlay 

4) A select menu is added to the header with the following options:
	- No status
	- Show Publication status
	- Show Visibility status (Icon in progress, use visibility/preview icon for now)
	- Show Permission status

We also display the number of items on the page that match the condition. This number is displayed as `pill` at the end of the option. It will help users by indicating which statuses match a condition.

- publication status should count modified and not published content
- visibility should count any content that has any visibility condition added
- permission should count any content that has any ACL overrides (e.g. changing live or edit roles)

If a content has been marked for deletion, then it should not be included in the count. Conversely, it should be added in the appropriate count when it is undeleted.

Selection is persisted across sites; It is also saved in browser local storage and should be restored accordingly.

Those choices are **NOT CUMULATIVE**, only one type of status can be checked at a time.

### Rules

**When Publication status is activated,** 
Use case 1: The page is published, and an orange border + badge (icon / label) is displayed around each item that is not published or modified.
Use case 2: The page is **not published**, then we don't check for children items
Use case 3: When a list/area/content with childNodes is not published, then we don't check for children items

**When Visibility status is activated** 
- time based or channel condition: orange border + badge
- Language condition: display with grey border and grey overlay + 'Visibility Condition' orange badge. When the language condition is removed, the item becomes visible again.

**When permission status is activated**
Display an orange border + badge around each item that has an ACL override permission (same rule as Page Composer)

Figma link: https://www.figma.com/design/kutQN9BvxduIuMDAGGFit5/Displaying-status-in-Page-Builder?node-id=2584-7196&t=9iaENFw9jn1BXMOM-0

gflores-jahia:
### Description

Implementation of PB content status story

- Implement `contentStatus.redux.jsx` for keeping track of current active status, and also keeping track of paths with given status categories (permissions, visibility conditions, and publication) to display count for a given page;
  - Also manages local storage for fetching and storing updates for current active status
- Implement and display content status dropdown and status count on jcontent header (using content status redux state)
- Implement `useBoxStatus` hook that contains all logic to display content status in `Box.jsx` and returns component to render badges
  - Tracks status badges to display depending on the status selected in the dropdown 
  - Provides `BoxStatus` component to render badges
  - Updates paths in status redux state for each status for a given node
- UI changes in `Box.jsx` to display status borders, badges and overlays 

Cypress tests to follow

#### Source code
- [x] I've considered the implications on security (in particular for changes to authentication, authorization, data fetching, ...)
- [x] I've considered the implications on performances
- [x] I've considered the implications on migration
- [x] I've considered the implications on code maintainability

#### Tests

#### Documentation
- [x] I've provided inline documentation (Source code)
- [ ] I've provided internal documentation (README, Confluence)
- [ ] I've provided user-facing documentation (Academy)

> [!TIP]


src/javascript/JContent/ContentRoute/ToolBar/ContentStatusSelector/ContentStatusSelector.jsx
```diff
@@ -0,0 +1,49 @@
...
+import styles from './ContentStatusSelector.scss';
+
+export const ContentStatusSelector = () => {
+    const {PUBLISHED, PERMISSIONS, VISIBILITY, NO_STATUS} = JContentConstants.statusView;
+    const statusMode = useSelector(state => state.jcontent.contentStatus.active) || NO_STATUS;
+    const contentStatus = useSelector(state => state.jcontent.contentStatus.statusPaths, shallowEqual);
+    const {t} = useTranslation('jcontent');
+    const dispatch = useDispatch();
+
+    const icons = {
```
r3dm1ke, line 16: could be outside of component?

r3dm1ke (approved):

src/javascript/JContent/ContentRoute/ToolBar/ContentStatusSelector/ContentStatusSelector.jsx
gflores-jahia: Will keep this one for now, but I'll make the change next time for content status-related PRs

AKarmanov (approved):
LGTM

PR #1696, opened 2025-03-26
chore: QABACKLOG-1585 : add config for cluster

ksagnard:


.github/workflows/nightly.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           module_id: jcontent
           incident_service: jcontent-${{ matrix.config_file }}
           jahia_image: ${{ matrix.jahia_image }}
+          ahia_cluster_enabled: false
```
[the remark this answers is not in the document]
ksagnard: fixed

sdusolle (dismissed):

gflores-jahia (commented):
Do we need to have some kind of check that makes sure all nodes are up before starting the tests ? 

Not sure if it needs to happen here or maybe in jahia-cypress.

.github/workflows/weekly-cluster.yml
```diff
@@ -0,0 +1,77 @@
+name: Nightly Test run
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  '0 0 * * 1'
```
rknj, line 6:
```suggestion
    - cron:  '0 0 * * 7'
```
Idea is to have it run on a day when the nightly will not run

ksagnard: done

PR #1705, opened 2025-03-28
fix: Add restriction check for structured view

### LINKED ISSUE #1678: JContent 3.2.1 does not respect area level node type restrictions

Jahia 8.2.0.6
JContent 3.2.1

### Steps to reproduce

1. Use the attached [eiffage.zip](https://github.com/user-attachments/files/19373904/eiffage.zip) template set (or use built jar renamed to zip attached to comment below https://github.com/user-attachments/files/19466875/eiffage-1.0.0-SNAPSHOT.zip)
2. Create a site with them (requires `legacy-default-components` module to be enabled on the site to activate jnt:banner)
3. On the home page, create a rich text in the pagecontent area
4. Use the menu to copy the rich text
5. You can now paste it inside the header absolute area, which is restricted to Banner objects

Image
Image
Image
Image
Image

### Current Behavior

- JContent allows pasting non-displayable node inside restricted areas
- This does not happen on content list, so the issue is most probably in the resolutions of constraints from the list to the area in the template

### Desired Behavior

JContent should not offer to paste if not allowed to create this type of node in the area

gflores-jahia:
### Description

Follow-up fix to https://github.com/Jahia/jcontent/pull/1694

- Revise implementation and store nodeType restrictions in JahiaAreasUtil instead to be able to use it across any jcontent view mode.
- Bumped `@jahia/cypress` to 4.3.0
- Added `data-node-name` selector to jcontent table row to be able to select by node name
  - added `contentTable.getRowByName` in cypress page object
- Added cypress test

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/ContentRoute/ContentLayout/ContentTable/Row.jsx
+++ b/src/javascript/JContent/ContentRoute/ContentLayout/ContentTable/Row.jsx
@@ -63,6 +63,7 @@ export const Row = ({
                       width: '100%'
                   }}
                   data-cm-role="table-content-list-row"
+                  data-node-name={node.name}
                   className={clsx(css.tableRow, (isCanDrop || isCanDropFile) && 'moonstone-drop_row', dragging && 'moonstone-drag')}
                   isHighlighted={isPreviewSelected}
                   onClick={() => {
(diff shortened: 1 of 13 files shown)


gflores-jahia:
Disabling content status test suite due to https://github.com/Jahia/visibility/issues/27

r3dm1ke (approved):

PR #1716, opened 2025-04-02
feat: Implement moonstone CardSelector

### LINKED ISSUE #1704: Implement CardSelector in Content Editor
- Implement our current component `CardSelector` in Content Editor: [ReferenceCard](https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/DesignSystem/ReferenceCard/ReferenceCard.jsx), [OrderableValue](https://github.com/Jahia/jcontent/tree/master/src/javascript/ContentEditor/DesignSystem/OrderableValue)
- We must avoid breaking changes in Content Editor, so you can update the component if needed
- Add actions on the Card to move the card first/up/down/last

abourgoindev:
### Description
Implement moonstone's CardSelector to replace ReferenceCard
Add order function in Picker
Add order buttons on ReferenceCard in OrderableValue for Picker


Eevolee:
Waiting for [CardSelector improvement](https://github.com/Jahia/moonstone/issues/988)

Eevolee:
[This ReferenceCard test](https://github.com/Jahia/jcontent/pull/1716/files#diff-40070061c7a2911409c3edec69dfa3024837ef01de04ee14ade7113b381eae93R146) doesn't work, I've tried various ways to find the target & call the onClick function

Eevolee:
The [id prop](https://github.com/Jahia/jcontent/pull/1716/files#diff-13b188cf6eb2e8f31565443dbca04c15c9693fa2e065c7e18f781e4ba4fd0642R71) is required for CardSelector but I set it as optional to avoid breaking changes. Should it be required in ReferenceCard?

Eevolee:
> > The [id prop](https://github.com/Jahia/jcontent/pull/1716/files#diff-13b188cf6eb2e8f31565443dbca04c15c9693fa2e065c7e18f781e4ba4fd0642R71) is required for CardSelector but I set it as optional to avoid breaking changes. Should it be required in ReferenceCard?
> 
> Is it really required for CardSelector or can that be made optional? if not, then we have field information we can pass as ID, or reuse the keys for the components

It is required for [this attribute](https://github.com/Jahia/moonstone/blob/develop/src/components/CardSelector/CardSelector.tsx#L77) on the CardSelector

src/javascript/ContentEditor/DesignSystem/OrderableValue/OrderableValue.jsx
```diff
@@ -32,6 +34,32 @@ export const OrderableValue = ({field, onFieldRemove, onValueReorder, index, com
                 <div className={styles.draggableCard}>
                     {component}
                 </div>
+                ) :
+                component.type === ReferenceCard ? (
```
gflores-jahia, line 38: I know this is still in progress but for this part, I think it might better/cleaner if we just implement the ReferenceCard here rather than taking it from component param. We could just pass in a flag e.g.`isReferenceCard` and create the ReferenceCard component instance here instead of passing then cloning.

Eevolee: I tried a different approach in [my latest commit](https://github.com/Jahia/jcontent/pull/1716/commits/8a121b4d01c9752a804b2e0438a9a2c5212353a6) by [setting the arrow buttons directly in Picker](https://github.com/Jahia/jcontent/pull/1716/commits/8a121b4d01c9752a804b2e0438a9a2c5212353a6#diff-9f3b30af5b4bd3a9b6a1fe98e54680112786f9de8727a85d9279ee97c4aaa87aR81) since ReferenceCard is already returned there, the only thing is that the display of `cardActions` on hover got a bit complicated and is now dispersed between [Picker's style](https://github.com/Jahia/jcontent/pull/1716/commits/8a121b4d01c9752a804b2e0438a9a2c5212353a6#diff-ce4b2c06536f4c13c1214827340e7265adc00d53712c2cf3dbf7e595c64cd336R28) & [OrderableValue's style](https://github.com/Jahia/jcontent/pull/1716/commits/8a121b4d01c9752a804b2e0438a9a2c5212353a6#diff-534ba8c6c7e788bc7a2add61dae6fcf5fcaee434361d08a2b429e61acc3494a6R35). There must be a way to simplify it.

gflores-jahia (commented):
~~Hey @Eevolee, looks like this branch have some conflicts with the recent merging work you've done, and might need to be synced up first with master.~~ Sorry it's my branch that needs to be synced up with your latest.

gflores-jahia (dismissed):
lgtm !

gflores-jahia (approved):

PR #1749, opened 2025-04-12
fix: Prevent proto pollution in json overrides

gflores-jahia:
### Description

https://support.jahia.com/browse/SECURITY-529

Prevent proto pollution by checking for reserved keywords before copy when merging json override objects.

In the case of set function, I've replaced it with utility package es-toolkit (more up-to-date lodash) that takes care of the corner cases around proto pollution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "connected-react-router": "^6.9.3",
     "dayjs": "^1.10.6",
     "ellipsize": "^0.5.0",
+    "es-toolkit": "^1.39.4",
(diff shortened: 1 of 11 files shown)


bpapez (dismissed):
Approving it as a quick solution, but using lodash would be indeed preferred.

bpapez (dismissed):

AKarmanov (approved):

sdusolle:
Cypress test is GREEN 
image

PR #1751, opened 2025-04-16
feat: Merge all drag and drop implementations

### LINKED ISSUE #1708: Use the same dnd implementation in Content Editor
**As an**  Editor
**I want to** have the same behavior when I re-order items
**In order to** a consistent experience

### Acceptance criteria

**Given** I edit a list item or a multiple picker
**When** I re-order items
**Then** I have the same experience

**Given** I edit a list item or a multiple picker
**Then** no drag icon is displayed
**And** a cursor `grab` is displayed when I hover the card to indicate the item is moveable
**And** the entire card can trigger the drag-n-drop

### Notes
We use two different implementations of re-ordering cards in a list:
- Multiple pickers: [Pickers.jsx](https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/SelectorTypes/Picker/Picker.jsx#L173-L193)
- Reordering a list [MultipleField.jsx](https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/MultipleField.jsx#L35-L55)

That is the first piece of code that can be refactored; you must also check if we can merge other things. For example, for the implementation of the list reordering, we insert a `div` between each item, but not for the multiple pickers.

I prefer to use the implementation of the Multiple Picker because it seems easier to use for users.

abourgoindev:
Replace function onValueReorder in Picker, Multiplefields & ManualOrdering by utils function onListReorder.


Eevolee:
Remaining issues:

- Need to redo the tests in `dragAndDrop.specs.js` in utils
- Returns to top of the page after reorder on a `ManualOrdering` component
- Inputs in `MultipleFields` loose focus after typing a single caracter

src/javascript/ContentEditor/utils/dragAndDrop.scss
```diff
@@ -1,28 +1,28 @@
...
-
-    text-transform: uppercase;
 }
 
 .referenceDropGhostHidden {
-    width: 100%;
-    height: 0;
-
-    background-color: var(--color-accent);
+    height: 8px;
```
gflores-jahia, line 12: [nitpick] could've maybe used global spacing var here

gflores-jahia (approved):
Thank you for the work ! I can see the handling of reorder looks better already. We should be ok to merge this.

But one thing to note here is that I still see some duplicated patterns in both `OrderableValue` and `DraggableReference` that I think we'll also be able to merge together as they look similar in implementation. Would love to hear your thoughts on this one.

But we can create a separate story for that one considering the work you already put into this one and sets jcontent up better to be able to do that.

gflores-jahia:
Btw feel free to skip sonarqube issue for now. lmk if sonarqube is still a blocker for you to merge, and I can help with that.

src/javascript/ContentEditor/utils/dragAndDrop.scss
Eevolee: This will be fixed in [this draft](https://github.com/Jahia/jcontent/pull/1716)

PR #1797, opened 2025-05-23
fix: prevent dnd when there is only one item in the list

### LINKED ISSUE #1790: Prevent drag and drop when there is only one item in a list
> [!IMPORTANT]
> - [ ] Before working on this issue, ensure it follows our [Definition of Ready](https://jahia-confluence.atlassian.net/wiki/spaces/PR/pages/2064566/Definition+of+Ready+-+Ref+ISPOL08.A140180)

**As a** user
**I want to** only have access to useful behavior
**In order to** have an interface that is easy to use

### Context
Since there is no reason to reorder a list with only one item, the drag-and-drop functionality should be disabled. 

### Acceptance criteria

**Given** Content editor with a multiple field
**When** there is only one item
**Then** the drag and drop is disabled

**Given** Content editor in Content List ordering
**When** there is only one item
**Then** the drag and drop is disabled

### Notes
- We can use the same logic than [useNodeDrag](https://github.com/Jahia/jcontent/blob/master/src/javascript/JContent/dnd/useNodeDrag.jsx#L30)

> [!IMPORTANT]
> - [ ] Before closing this issue, ensure it follows our [Definition of Done](https://jahia-confluence.atlassian.net/wiki/spaces/PR/pages/2072202/Definition+of+Done+-+Ref+ISPOL08.A14030)

ffffffelix:
### Description
Prevent dnd when there is only one item in a list

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/ContentEditor/DesignSystem/OrderableValue/OrderableValue.jsx
+++ b/src/javascript/ContentEditor/DesignSystem/OrderableValue/OrderableValue.jsx
@@ -6,6 +6,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 export const OrderableValue = ({field, onFieldRemove, onValueReorder, index, component, isReferenceCard}) => {
+    const isDraggable = component ? component.props.isDraggable : false;
     const {t} = useTranslation('jcontent');
     const name = `${field.name}[${index}]`;
     const [{isDropping}, drop] = useDrop({
@@ -16,6 +17,7 @@ export const OrderableValue = ({field, onFieldRemove, onValueReorder, index, com
         }
     });
     const [{isDragging}, drag] = useDrag({
+        canDrag: () => isDraggable,
         type: `REFERENCE_CARD_${field.name}`, item: {name: name}, collect: monitor => ({
             isDragging: monitor.isDragging()
         })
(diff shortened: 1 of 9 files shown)


gflores-jahia (dismissed):

src/javascript/ContentEditor/DesignSystem/OrderableValue/OrderableValue.jsx
```diff
@@ -1,24 +1,47 @@
...
+    if (typeof component.props.readOnly === 'boolean') {
+        return !component.props.readOnly;
+    }
+
+    if (typeof component.props.isDraggable === 'boolean') {
+        return component.props.isDraggable;
+    }
+
+    // Check if the component has multiple values
+    return (component.props.values?.[component.props.field?.name]?.length ?? 0) > 1;
```
gflores-jahia, line 23: should we evaluate this first, i.e. prevent if component has one value but marked as isDraggable?

src/javascript/ContentEditor/DesignSystem/OrderableValue/OrderableValue.jsx
```diff
@@ -31,11 +37,11 @@ export const OrderableValue = ({field, onFieldRemove, onValueReorder, index, com
         >
             <div className={`${styles.referenceDropGhostHidden} ${isDropping ? styles.referenceDropGhost : ''}`} data-droppable-zone={name}/>
             {component &&
-                <div className={styles.draggableCard} {...(isReferenceCard && {ref: drag})}>
+                <div ref={isReferenceCard && isDraggable ? drag : null} className={styles.draggableCard}>
```
gflores-jahia, line 40: `react-dnd` adds an attribute `draggable` that does not reflect whether an element can indeed be dragged or not from `canDrag` so it's not best to depend on react-dnd here. So we add drag ref only if we know element is draggable or not.

https://github.com/react-dnd/react-dnd/issues/2690

gflores-jahia (approved):

PR #1812, opened 2025-05-30
fix: improved dnd behaviour in pickers & content lists

r3dm1ke:
### Description
This PR rewrites the DnD logic in jContent. The experience is improved and easier to use.
The change targets orderable content (list of links, list of references) and multiple fields (pickers).
The way this is done is using a new custom hook, `useReorderList`. It saves a copy of items, and exposes functions to reorder them in place. Each item becomes a drop target, and will call this method to insert a draggable item either above or below itself.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/ContentEditor/utils/useReorderList.js
+++ b/src/javascript/ContentEditor/utils/useReorderList.js
@@ -60,7 +60,7 @@ export function useReorderList(items) {
         }
 
         setReorderedItems(newReorderedItems);
-        itemsRef.current = {items, serialized: JSON.stringify(items)};
+        itemsRef.current = {items, serialized: currentItemsSerialized};
         return {reorderedItems: newReorderedItems, handleReorder, reset};
     }
 
@@ -73,6 +73,7 @@ export function useReorderList(items) {
 
 /**
  * Custom hook to handle the drop event for reordering items.
+ * This hook main purpose is to provide a reusable hover function, designed to work with useReorderList
  * @param {object} args    object contains the following properties:
  * @param {object} args.ref    reference to the drop target
  * @param {number} args.index    index of the item being hovered
--- a/src/javascript/ContentEditor/utils/dragAndDrop.scss
+++ b/src/javascript/ContentEditor/utils/dragAndDrop.scss
@@ -16,12 +16,12 @@
     pointer-events: none;
 }
 
-.draggingCard {
-    background-color: var(--color-accent_light_plain20) !important;
+.draggableCard.draggingCard {
+    background-color: var(--color-accent_light_plain20);
 }
 
-.draggingCard * {
-    visibility: hidden !important;
+.draggableCard.draggingCard * {
+    visibility: hidden;
 }
 
 .draggableCard {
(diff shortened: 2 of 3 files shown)


src/javascript/ContentEditor/utils/useReorderList.js
```diff
@@ -0,0 +1,165 @@
...
+         */
+        let newReorderedItems;
+        if (itemsRef.current.items.length === items.length) {
+            newReorderedItems = updateDragItems(items, reorderedItems);
+        } else {
+            newReorderedItems = makeDragItems(items);
+        }
+
+        setReorderedItems(newReorderedItems);
+        itemsRef.current = {items, serialized: JSON.stringify(items)};
```
gflores-jahia, line 63: we can reuse `currentItemsSerialized`

src/javascript/ContentEditor/utils/useReorderList.js
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Custom hook to handle the drop event for reordering items.
+ * @param {object} args    object contains the following properties:
+ * @param {object} args.ref    reference to the drop target
+ * @param {number} args.index    index of the item being hovered
+ * @param {function} args.onReorder    function to call when the item is reordered during hover
+ * @param {object} useDropArgs    additional arguments for the useDrop hook
+ */
+export function useReorderDrop(args, useDropArgs) {
+    const hover = (item, monitor) => {
```
gflores-jahia, line 83: I'm guessing the whole purpose of the custom hook for drop is to modify the hover ? Can we explain a bit what we're exactly doing here as a comment

src/javascript/ContentEditor/utils/useReorderList.js
```diff
@@ -0,0 +1,165 @@
...
+        let newReorderedItems;
+        if (itemsRef.current.items.length === items.length) {
+            newReorderedItems = updateDragItems(items, reorderedItems);
+        } else {
+            newReorderedItems = makeDragItems(items);
+        }
+
+        setReorderedItems(newReorderedItems);
+        itemsRef.current = {items, serialized: JSON.stringify(items)};
+        return {reorderedItems: newReorderedItems, handleReorder, reset};
```
gflores-jahia, line 64: is return here necessary ? wouldn't the set call on line 62 trigger the return call on line 67 with new value ?

gflores-jahia (commented):
Also PR could use a bit of overall description as to what was changed.

gflores-jahia (approved):

src/javascript/ContentEditor/DesignSystem/OrderableValue/OrderableValue.jsx
```diff
@@ -1,71 +1,72 @@
...
+                className={clsx(
+                    styles.draggableCard,
+                    isDragging && styles.draggingCard
+                )}
+                data-sel-content-editor-multiple-generic-field={name}
+                data-sel-content-editor-field-readonly={field.readOnly}
+                data-handler-id={handlerId}
+            >
+                {!isReferenceCard &&
+                    <div ref={isDraggable ? drag : null} className={clsx(isDraggable ? styles.draggableIcon : styles.notDraggableIcon)}>
```
gflores-jahia, line 44: just want to confirm change in logic here, as it looks like we now lose drag ref when it's draggable and isReferenceCard

gflores-jahia (approved):

PR #1829, opened 2025-06-05
test: QABACKLOG-1547: import export special characters in jcontent

ksagnard:


sdusolle (approved):

gflores-jahia (approved):
looks good, thank you for the ExportDialog refactor ! just need minor change with the default selector (see below)

tests/cypress/page-object/exportDialog.ts
```diff
@@ -0,0 +1,19 @@
+import {BaseComponent, Button, Dropdown, getComponentByAttr} from '@jahia/cypress';
+
+export class ExportDialog extends BaseComponent {
+    static defaultSelector = 'button[data-sel-role="breadcrumb-item"]';
+
+    export(downloadsFolder:string, workspace:string = 'Staging content only', exportFormat: string = 'zip') {
+        const dialog = getComponentByAttr(BaseComponent, 'data-cm-role', 'export-options');
+        dialog.should('be.visible');
```
gflores-jahia, line 8: no need to fetch dialog here as this class should be the export dialog. this selector should be changedas default selector instead.

gflores-jahia (changes requested):

tests/cypress/page-object/exportDialog.ts
```diff
@@ -1,19 +1,19 @@
 import {BaseComponent, Button, Dropdown, getComponentByAttr} from '@jahia/cypress';
 
 export class ExportDialog extends BaseComponent {
-    static defaultSelector = 'button[data-sel-role="breadcrumb-item"]';
+    static defaultSelector = '[data-cm-role="export-options"]';
 
     export(downloadsFolder:string, workspace:string = 'Staging content only', exportFormat: string = 'zip') {
         const dialog = getComponentByAttr(BaseComponent, 'data-cm-role', 'export-options');
```
gflores-jahia, line 7: sorry one more minor thing, but I think this line should be safe to remove. 

I don't think we use `dialog` anywhere now.

ksagnard: we use it here  getComponentByAttr(Dropdown, 'data-cm-role', 'select-workspace', dialog).select(workspace); 
(line 10)

gflores-jahia: ah, thanks. in this case `getComponentByAttr` has `parent` param of type `BaseComponent`. 

Here, we can pass `this` instead of `dialog` - `getComponentByAttr(Dropdown, 'data-cm-role', 'select-workspace', this).select(workspace);`

tests/cypress/page-object/exportDialog.ts
```diff
@@ -4,10 +4,10 @@ export class ExportDialog extends BaseComponent {
     static defaultSelector = '[data-cm-role="export-options"]';
 
     export(downloadsFolder:string, workspace:string = 'Staging content only', exportFormat: string = 'zip') {
-        const dialog = getComponentByAttr(BaseComponent, 'data-cm-role', 'export-options');
+        getComponentByAttr(BaseComponent, 'data-cm-role', 'export-options');
```
gflores-jahia, line 7: ok to remove this full line as it's not used anymore.

ksagnard: done

gflores-jahia (approved):

PR #1849, opened 2025-06-19
fix: add terms limit and debounce for picker search

AKarmanov:
### Description
Add terms limit and debounce for picker search this prevents searching on one or two letters which can result in a very large result and limits the number of queries submitted.

#### Source code

#### Tests

> [!TIP]


gflores-jahia (approved):
looks ok to me.

PR #1856, opened 2025-06-25
fix: Clear clicked element selection when copy action is used

AKarmanov:
### Description
Clear clicked element selection when copy action is used, this allows to see paste buttons right away

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/EditFrame/EditFrame.jsx
+++ b/src/javascript/JContent/EditFrame/EditFrame.jsx
@@ -218,6 +218,7 @@ export const EditFrame = () => {
         }, 50);
 
         return () => {
+            setClickedElementHook(() => undefined)
             clearInterval(interval);
         };
     }, []);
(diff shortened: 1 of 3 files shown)


r3dm1ke (dismissed):

gflores-jahia (commented):
We should maybe check if we have other options on how to share states in a React manner across the components. I added comment below.

src/javascript/JContent/EditFrame/EditFrame.jsx
```diff
@@ -84,6 +95,7 @@ export const EditFrame = () => {
     const [currentUrlParams, setCurrentUrlParams] = useState('');
     const [previousUrlParams, setPreviousUrlParams] = useState('');
     const [clickedElement, setClickedElement] = useState();
+    setClickedElementHook(setClickedElement);
```
gflores-jahia, line 98: We should also robably add a clean up on unmount

gflores-jahia (approved):

PR #1857, opened 2025-06-25
fix: Add check for multiple keyword on tag selector type

### LINKED ISSUE #1822: Content editor crashes when declaring a property string,tag without the keyword "multiple"

8.2.1.0 (standard installation)

### Steps to reproduce

Add a custom module with definition
  
```
[jnt:wording] > jnt:content, jmix:basicContent, jmix:editorialContent
  - text (string) i18n
  - version (string,tag)  
```
And activate the module. 

### Current Behavior

Add a content of the type wording, and add two tags into the field and save the content. Content is correctly saved, but when you try to reopen the editor it crashes

### Desired Behavior

Editor shouldn't crash.

gflores-jahia:
### Description

Add an error message when Tag definition doesn't have multiple keyword as we do not support single tags at the moment.

We do not display message through validation errors as we want to distinguish this from a user-generated error, and validation errors prevents user from saving content.

Added cypress test.

#### Source code
- ~[ ] I've shared and documented any breaking change~
- ~[ ] I've reviewed and updated the jahia-depends~

#### Tests

> [!TIP]

### DIFF
--- a/tests/cypress/e2e/menuActions/delete.cy.ts
+++ b/tests/cypress/e2e/menuActions/delete.cy.ts
@@ -32,7 +32,7 @@ describe('delete tests', () => {
         jcontent.getAccordionItem('pages').getTreeItem('test-pageDelete1').contextMenu().select('Delete');
 
         cy.log('Verify dialog opens and can be mark for deletion');
-        getComponent(DeleteDialog).markForDeletion('You are about to delete 5 items, including 3 page(s)');
+        getComponent(DeleteDialog).toggleRowExpanded().markForDeletion('You are about to delete 5 items, including 3 page(s)');
 
         cy.log('Verify menu and subpages has been marked for deletion');
         cy.apollo({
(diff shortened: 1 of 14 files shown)


tests/cypress/e2e/menuActions/delete.cy.ts
```diff
@@ -32,7 +32,7 @@ describe('delete tests', () => {
         jcontent.getAccordionItem('pages').getTreeItem('test-pageDelete1').contextMenu().select('Delete');
 
         cy.log('Verify dialog opens and can be mark for deletion');
-        getComponent(DeleteDialog).markForDeletion('You are about to delete 5 items, including 3 page(s)');
+        getComponent(DeleteDialog).toggleRowExpanded().markForDeletion('You are about to delete 5 items, including 3 page(s)');
```
gflores-jahia, line 35: I added toggleRow for debugging as previous runs was showing 6 items instead of 5. Not able to reproduce anymore, but keeping for next time. 

Earliest instance of failure I see so far was from this run https://github.com/Jahia/jcontent/actions/runs/15856427628/job/44703255994

r3dm1ke (approved):

AKarmanov (approved):

PR #1872, opened 2025-07-07
feat: added config option for CE header action buttons

r3dm1ke:
### Description
This PR adds a new config option `contentEditorQuickActionsButtons.limit` to jContent that controls the number of quick action buttons in CE editor.

#### Source code

#### Tests

> [!TIP]


src/javascript/ContentEditor/ContentEditor/EditPanel/HeaderActions/utils.js
```diff
@@ -0,0 +1,16 @@
+import {registry} from '@jahia/ui-extender';
+
+export const getButtonLimitValue = () => {
+    const configValue = contextJsParameters.config.jcontent?.['contentEditorQuickActionsButtons.limit'];
+    const parsedValue = parseInt(configValue, 10);
```
gflores-jahia, line 5: default limit value from config different than default here? Maybe make it the same for consistency

gflores-jahia: Also if we can rename utils to utils.headerActions.js instead.

src/main/resources/META-INF/configurations/org.jahia.modules.jcontent.cfg
```diff
@@ -11,3 +11,5 @@ createChildrenDirectButtons.limit = 5
 # forceDateFormat overrides the browser locale format of date in content editor
 # Possible values are: DD/MM/YYYY, MM/DD/YYYY
 # forceDateFormat=DD/MM/YYYY
+
+contentEditorQuickActionsButtons.limit=5
```
gflores-jahia, line 15: I would actually just leave this commented out and with explanation of what this does (similar to what we have for forceDateFormat above).

gflores-jahia (changes requested):
Don't we also need to provide migration script to add

gflores-jahia (approved):

PR #1879, opened 2025-07-10
fix: Remove context menu in page builder completely.

### LINKED ISSUE #1878: Remove right click entry on pages
In page builder (visual mode) when no item is selected, by default the page itself is selected. Therefore if you do a right click, you'll have a menu that operates on the page item. In content folders and asset folders we had the same behavior that we removed when we cleaned the menus, because it was sometimes confusing and could led users to do mistakes.

it's also causing troubles when the user do a right click too fast while the entire page is not loaded. the menu is related to the page and not to the item the user think having targeted.

### Story
As an editor I don't want to access right click when I'm not sure on what it will apply.

### Note
We already have 3 dots and right click menu in the page tree + header buttons in the main panel to execute actions on the page, removing the central right-click will not penalize editors

## Testable scenarios

**Setup**
Given i'm in page builder

**Case 1**
When I do a right click on the page background
Then nothing happen

### Test Strategy
Cypress tests?

### Labels 
En: 
Fr: 
De:

AKarmanov:
### Description
Remove context menu in page builder completely. If I read #1878  correctly we want to remove context menu completely, if that is the case then provided solution should do the trick. 

#### Source code

#### Tests

> [!TIP]


gflores-jahia (approved):
lgtm, but can we maybe add like a @deprecate note with a link to this PR or something in BoxesContextMenu that it needs to be removed eventually ?

PR #1903, opened 2025-07-25
fix: Translation editor fixes

### LINKED ISSUE #1904: side-by-side translation UI
- PC feature parity
- Provide a usable an efficient user experience to translate from one language to another
- Mockup: [Figma](https://www.figma.com/design/xKdfkN9Pe9JRbD9BfMXKAY/Translate-screens?node-id=2856-30939&t=7Iadc9TyUEzlQHuR-1)

### Story

Add a "Translate" item in right-click and 3dots menus on page (page-tree/header) and on content items (rc / 3dots)
The translate item opens the sbs screen (use the full space available, like "advanced editing")
By default the sbs screen opens with the current language as "source" and the first default language as "destination"
The SBS screen displays the "Content" section fields only + the SEO section if it's a Page or a MainContent
In front of each property (field) an arrow allows to copy the value to the other side
When scrolling, both sides scroll at once
The source language is read-only (not editable)
The tranlator can change the source and the destination language
If the translator change destination language before saving
-- ideally we keep in memory the changes that wil be saved at the end of the session
-- we draw a warning ("Changes in {language} has not been saved, do you want to save them before changing language")

Save button = just a save, do not close
Close = verify if unsaved changes. if yes = call the modal we use in CE, if not just close
Publication button: can we remove "publication manager" entry from this menu ?
Title : tile of the content we are editing, if no title exist yet, then system name (CE default behavior) or "missing title" label
Status : display all statuses badges EXCEPT publication (highly misleading here)
Content section only
Content sectionand SEO sections if jnt:page or jnt:maincontent
Sections are "open" (expanded) by default
3 dots on the editing side : hidden if empty. We do not provide any action by default
Blue arrow : copy the field value to the opposite language
Blue arrow is centered on the field height 
A blue bar is displayed on the full height of the field, on hover of the lue arrow

## Testable scenarios

**Setup**
Given

**Case 1**
When 
Then 

**Case N**
When 
Then 

### Test Strategy
Cypress tests?

### Labels 
En: 
Fr: 
De:

gflores-jahia:
### Description

Follow-up fixes from PR review:

- Rename `sbsContext` to `sideBySideContext`
- Add translations for translation editor language dropdowns

### DIFF
--- a/src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelLanguageSwitcher/EditPanelLanguageSwitcher.spec.jsx
+++ b/src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelLanguageSwitcher/EditPanelLanguageSwitcher.spec.jsx
@@ -32,7 +32,7 @@ jest.mock('~/ContentEditor/contexts/ContentEditorConfig/ContentEditorConfig.cont
     return {
         useContentEditorConfigContext: () => ({
             lang: 'fr',
-            sbsContext: {}
+            sideBySideContext: {}
         })
     };
 });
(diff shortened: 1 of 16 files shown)


AKarmanov (approved):

rknj:
Fixes https://github.com/Jahia/jcontent/issues/1904

PR #1929, opened 2025-07-31
fix: Hide persona panel on compare staging/live dialog

Janin-Michel-Mathias:
### Description
Added url parameter hidePersonaPanel to url of the staging/live left iframe's ref

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/CompareDialog/FrameManager/FrameManager.jsx
+++ b/src/javascript/JContent/CompareDialog/FrameManager/FrameManager.jsx
@@ -65,7 +65,7 @@ const FrameManager = () => {
     // Load frames
     useEffect(() => {
         if (leftFrame.current && data?.jcr?.result?.renderUrlEdit) {
-            leftFrame.current.contentWindow.location.href = resolveUrlForLiveOrPreview(data.jcr.result.renderUrlEdit, false);
+            leftFrame.current.contentWindow.location.href = resolveUrlForLiveOrPreview(data.jcr.result.renderUrlEdit, false, null, true);
         }
 
         if (rightFrame.current && data?.jcr?.result?.renderUrlLive) {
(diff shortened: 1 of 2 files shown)


src/javascript/JContent/CompareDialog/FrameManager/FrameManager.jsx
```diff
@@ -65,7 +65,7 @@ const FrameManager = () => {
     // Load frames
     useEffect(() => {
         if (leftFrame.current && data?.jcr?.result?.renderUrlEdit) {
-            leftFrame.current.contentWindow.location.href = resolveUrlForLiveOrPreview(data.jcr.result.renderUrlEdit, false);
+            leftFrame.current.contentWindow.location.href = resolveUrlForLiveOrPreview(data.jcr.result.renderUrlEdit, false, null, true);
```
jsinovassin, line 68: I would not have modified the resolveUrlForLiveOrPreview function. 
I would just have added hidePersonaPanel=true at the end url like it's done in the customizedPreviewApp.jsx file
Example: resolveUrlForLiveOrPreview(renderUrl)}${urlParams}`
Here the modification would be:
 leftFrame.current.contentWindow.location.href = `${resolveUrlForLiveOrPreview(renderUrl)}?hidePersonaPanel=true`
And it would be nice to add a comment to explain that the hidePersonaPanel is related to jExperience and that modification is necessary to hide the persona panel in edit in the comparison page

jsinovassin (changes requested):

src/javascript/JContent/CompareDialog/FrameManager/FrameManager.jsx
```diff
@@ -65,7 +65,7 @@ const FrameManager = () => {
     // Load frames
     useEffect(() => {
         if (leftFrame.current && data?.jcr?.result?.renderUrlEdit) {
-            leftFrame.current.contentWindow.location.href = resolveUrlForLiveOrPreview(data.jcr.result.renderUrlEdit, false);
+            leftFrame.current.contentWindow.location.href = `${resolveUrlForLiveOrPreview(data.jcr.result.renderUrlEdit, false)}?hidePersonaPanel`;
```
jsinovassin, line 68: I think a comment here to explain why hidePersonaPanel is added here would help

jsinovassin (changes requested):

jsinovassin (approved):

PR #1955, opened 2025-08-09
fix: Display individual create buttons

AKarmanov:
### Description
When we display create buttons in GWT the process is slightly different, we never supply multiple types at once, which can cause content tree to change altering its roots. This means that in place of `jnt:contentReference` we will have `jmix:basicContent`, which is inconsistent with our expectations. This modifications makes sure that we mimic GWT behaviour when we have multiple node type restrictions.  

#### Source code

#### Tests

> [!TIP]


gflores-jahia (approved):
cypress test?

PR #1978, opened 2025-08-18
fix: check function existing to avoid error in browser console

jsinovassin:
### Description
Check if onMouseEnter is defined in MenuItemRenderer before calling it to avoid to have error in the browser console

#### Source code

#### Tests

> [!TIP]


gflores-jahia (approved):

brigitteradl:
First reproduced with jContent 3.3.1 -> in Chrome only and when opening several contextual menus the message appeared at some point in the console.

With jcontent-3.4.0-20250819.123200-95 I could not reproduce - OK

Issue #1985, opened 2025-08-20
testCreateMenuForReviewer


sdusolle:
Same scenario as [testCreateMenuForEditor](https://github.com/Jahia/jahia-private/issues/3250) but just login with `reviewer` instead

sdusolle:
We can shorten the test with : 
```
create one folder / a simpleList / a constraintList3

- login as reviewer
- open jcontent > content folders > contentEditorCreateMenuTest
- assert actions "new content" and "new content folder" are not displayed in the header

- open contextual menu for content list "simpleList"
- assert action "new content" is not displayed in the menu

- open contextual menu for list "constraintList3"
- assert actions "new constraintChild1", "new constraintChild2", "new constraintChild3" are not listed in the menu
```

PR #1995, opened 2025-08-22
test: add test for reviewer

sdusolle:
### Description
Convert Selenium test testCreateMenuForReviewer into a Cypress test
https://github.com/Jahia/jcontent/issues/1985

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/tests/cypress/e2e/menuActions/reviewer.cy.ts
+++ b/tests/cypress/e2e/menuActions/reviewer.cy.ts
@@ -1,0 +1,52 @@
+import {JContent} from '../../page-object';
+import {addNode, enableModule} from '@jahia/cypress';
+
+describe('Reviewer permissions', () => {
+    let jcontent: JContent;
+
+    before(() => {
+        enableModule('qa-module', 'digitall');
+        addNode({
+            name: 'my-list',
+            parentPathOrId: '/sites/digitall/contents',
+            primaryNodeType: 'jnt:contentList'
+        });
+        addNode({
+            name: 'constraintlist3',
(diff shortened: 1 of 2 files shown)


tests/cypress/e2e/menuActions/reviewer.cy.ts
```diff
@@ -0,0 +1,52 @@
...
+        // Check new content folder action is not displayed
+        cy.get('div[role="toolbar"]').find('button[data-sel-role="createContentFolder"]').should('not.exist');
+        cy.get('div[role="toolbar"]').find('button[data-sel-role="createContent"]').should('not.exist');
+
+        // Check action 'new content' under my-list is not displayed
+        const contextMenu = jcontent.openContextMenuByRowName('my-list');
+        contextMenu.shouldNotHaveItem('New content');
+
+        // Close contextual menu
+        cy.get('.moonstone-menu_overlay').click({force: true});
```
gflores-jahia, line 39: contextMenu has a close() function

gflores-jahia (changes requested):
if possible, it's preferable to create/delete a new site for every new test instead of using digitall (in case we would want to move away from digitall or becomes deprecated). Also makes sure no side effects for other tests as well.

We have the `createSite` helper function, and also `createUser` and `grantRoles` helper functions for the reviewer part (see [here](https://github.com/Jahia/jcontent/blob/master/tests/cypress/e2e/contentEditor/copyFromLanguageAction.cy.ts#L27-L28) for example)

PR #2051, opened 2025-11-14
feat: Add sorting in thumbnails mode

AKarmanov:
### Description
Add sorting in thumbnails mode

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/ContentRoute/ToolBar/SortSelector/SortSelector.jsx
+++ b/src/javascript/JContent/ContentRoute/ToolBar/SortSelector/SortSelector.jsx
@@ -10,23 +10,38 @@ import JContentConstants from '~/JContent/JContent.constants';
 const SORT_DATA = [
     {
         label: 'jcontent:label.contentManager.listColumns.name',
-        orderBy: 'displayName'
+        orderBy: 'displayName',
+        attributes: {
+            'data-sel-role': 'sel-media-sort-property-name'
+        }
     },
     {
         label: 'jcontent:label.contentManager.listColumns.type',
-        orderBy: 'content.mimeType.value'
+        orderBy: 'content.mimeType.value',
+        attributes: {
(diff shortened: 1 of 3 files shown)


src/javascript/JContent/ContentRoute/ToolBar/SortSelector/SortSelector.jsx
```diff
@@ -0,0 +1,113 @@
...
+            className={classes.label}
+            label={t('jcontent:label.contentManager.sortSelector.direction')}
+            variant="title"
+        />
+            <Dropdown
+                data-sel-role="sel-media-sort-order-dropdown"
+                variant="outlined"
+                value={sort?.order}
+                data={[
+                {
```
gflores-jahia: we should also insert selector roles for each of the dropdown items using `attributes` data object (see `ContentStatusSelector` for example) and use it for the test selectors instead of using labels.

src/javascript/JContent/ContentRoute/ToolBar/SortSelector/SortSelector.jsx
```diff
@@ -0,0 +1,113 @@
+import React, {useMemo} from 'react';
+import PropTypes from 'prop-types';
+import {shallowEqual, useDispatch, useSelector} from 'react-redux';
+import {useTranslation} from 'react-i18next';
+import {cmSetSort} from '~/JContent/redux/sort.redux';
+import {ArrowDown, ArrowUp, Dropdown, CustomDropdown, Separator, MenuItem} from '@jahia/moonstone';
+import classes from './SortSelector.scss';
+import JContentConstants from '~/JContent/JContent.constants';
+
+const SORT_DATA = [
```
gflores-jahia: we should also insert selector roles for each of the dropdown items using `attributes` data object (see `ContentStatusSelector` for example) and use it for the test selectors instead of using labels.

PR #2081, opened 2025-12-08
feat: Replace react pdf library

AKarmanov:
### Description
Replace react pdf library to avoid vulnerability.

#### Source code

#### Tests

> [!TIP]


gflores-jahia (approved):
Checked difference with existing React.lazy. it looks like the bulk of the pdf worker is already lazy loaded as a separate bundle. So we should be good.

PR #2101, opened 2025-12-12
fix: Update translation and update box display rules for referenced content

AKarmanov:
### Description
Update translation and update box display rules by avoiding displaying boxes for external content.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/EditFrame/Boxes/Boxes.jsx
+++ b/src/javascript/JContent/EditFrame/Boxes/Boxes.jsx
@@ -410,6 +410,7 @@ export const Boxes = ({currentDocument, currentFrameRef, currentDndInfo, addInte
                 </div>
             ));
     }, [
+        path,
         clickedElement,
         placeholders,
         nodes,


r3dm1ke (approved):

gflores-jahia (commented):
Nice one, that's a good filter to add for filtering out references.

Can we add a cypress test for this new behaviour of filtering out references? I feel we should've added it at the beginning with https://github.com/Jahia/jcontent/pull/2090.

Let me know if you even want the test on a separate PR and I can approve this for now.

AKarmanov:
> Nice one, that's a good filter to add for filtering out references.
> 
> Can we add a cypress test for this new behaviour of filtering out references? I feel we should've added it at the beginning with #2090.
> 
> Let me know if you even want the test on a separate PR and I can approve this for now.

We didn't know this issue existed on #2090 as specific content works fine. I feel like we should do a separate ticket for that, not sure how easy it would be to check that box situation with cypress, could require some time.

PR #2265, opened 2026-04-01
feat: replaces the legacy Tag Manager UI in jContent

smonier:
### Description
This PR replaces the legacy Tag Manager UI in `jcontent` with a new React 18 admin app integrated in the Additional Apps area, using Moonstone components and the current Jahia UI patterns.

Main changes:
- Added a new `TagManager` app in `jcontent`, registered as an admin route under Additional Apps.
- Replaced the old Webflow-based experience with a native React UI.
- Kept the Tag Manager entry in the same logical place in `jcontent`.
- Implemented full i18n coverage for the new UI labels and dialogs.
- Followed Moonstone/Jahia design patterns for:
  - sticky full-width header
  - table/list layout
  - dialogs
  - right-side usages panel
- Added the main Tag Manager features:
  - tag inventory table
  - search
  - client-side sorting
  - pagination
  - bulk rename
  - bulk delete
  - view usages in a right-side panel
  - rename a tag on a specific content
  - remove a tag from a specific content with confirmation
- Integrated the UI with the new `graphql-core` Tag Manager backend:
  - `tags(...)`
  - `taggedContent(...)`
  - `renameTag(...)`
  - `deleteTag(...)`
  - `deleteTagOnNode(...)`
  - `renameTagOnNode(...)`
- Removed the previous direct `jcr.mutateNode(...).setValues(...)` write path for per-content tag editing and switched that flow to the dedicated Tag Manager GraphQL mutation.
- Added explicit stable `data-cm-role` selectors to make the new UI testable and less dependent on visual/text selectors.

Frontend structure changes:
- Added a dedicated `src/javascript/TagManager/` area to isolate the new feature.
- Split the feature into focused components:
  - page container
  - table
  - right-side panel
  - dialogs
  - GraphQL queries/mutations
  - shared tag suggestion hook
- Extracted duplicated tag suggestion logic into a reusable hook.
- Reduced unnecessary UI refresh fan-out after mutations by replacing broad refresh behavior with more targeted refreshes.

Behavior/UI details implemented:
- header shows the current site name
- content names in the usages panel use the selected content language
- content type display name is shown in the usages panel
- destructive actions use `DeletePermanently` with danger styling
- dialog copy was aligned with the requested wording
- tag names are quoted in dialogs and side panel title
- side panel is rendered inside the content wrapper, anchored on the right
- clicking outside the side panel closes it
- selected tag row is highlighted when the usages panel is open

Test coverage:
- Added a Cypress E2E spec for Tag Manager covering:
  - registration in Additional Apps
  - opening the Tag Manager route
  - usages panel open/close behavior
  - click-outside close on the side panel
  - rename on a single content
  - bulk rename
  - bulk delete
  - special-character tag regression
- The special-character regression explicitly covers:
  - `view usages`
  - `rename`
  - `delete`
  for tags such as `ù^$ùç_"(_çt"à'çr(`

Notes:
- No `package.json` change was made.
- The current table search/sort/pagination behavior remains frontend-driven for now; server-side search is still a possible follow-up if needed for larger tag inventories.

#### Source code
- [x] I've shared and documented any breaking change
- [x] I've reviewed and updated the jahia-depends

#### Tests

Manual validations recommended:
- open Tag Manager from Additional Apps
- verify site header title
- verify usages panel opens and closes correctly
- verify per-content rename updates the targeted content only
- verify bulk rename/delete update the table correctly
- verify selected language is used for content display names in the usages panel
- verify a tag with special characters can be viewed, renamed, and deleted
- verify live/edit behavior with the backend deployed


smonier:
### Description
This PR adds the new native Tag Manager application to `jcontent` and replaces the legacy Tag Manager user experience with a React 18 + Moonstone implementation integrated in Additional Apps.

Main changes:
- Added a dedicated `TagManager` feature area under `src/javascript/TagManager/`.
- Registered Tag Manager as an Additional Apps admin route in `jcontent`.
- Added the `tagManager` permission definition to `jcontent`, which now owns permission exposure for the new UI.
- Implemented the new Tag Manager UI with:
  - sticky full-width header with site name
  - tag inventory table
  - search
  - sorting
  - pagination
  - bulk rename
  - bulk delete
  - usages side panel
  - per-content rename
  - per-content delete with confirmation
- Added full i18n coverage for the new UI.
- Added stable `data-cm-role` selectors for automated testing.

Integration changes:
- The UI uses the new Tag Manager GraphQL backend for all tag-management flows.
- Removed the previous direct `jcr.mutateNode(...).setValues(...)` write path for per-content tag updates.
- Notification counts were corrected to represent impacted content items rather than summing `EDIT` and `LIVE` workspace executions.

Behavioral fixes included during implementation:
- side panel remains within the content wrapper
- click-away closes the side panel
- content display names use the selected content language
- content type display name is shown in the usages panel
- special-character tags can be viewed, renamed, and deleted
- destructive actions use the requested Moonstone styling
- dialog copy and layout were aligned with the requested UX

Test work:
- Added a Cypress Tag Manager spec.
- Refactored the Cypress coverage to use a dedicated Tag Manager page object.
- Covered:
  - opening Tag Manager from Additional Apps
  - special-character tag regression (`view usages`, `rename`, `delete`)
  - bulk rename/delete
  - per-content rename
  - per-content delete with confirmation
  - click-away close on the side panel

Notes:
- No change was made to `package.json`.
- The current tag-list search/sort/pagination behavior remains frontend-driven.

### Checklist
#### Source code
- [x] I've shared and documented any breaking change
- [x] I've reviewed and updated the jahia-depends

#### Tests

src/javascript/TagManager/register.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import {Tag} from '@jahia/moonstone';
+import {TagManager} from './TagManager';
+
+export const register = registry => {
+    registry.add('adminRoute', 'tagsmanager', {
```
gflores-jahia: I think we need a different name as it conflicts with existing `tagsmanager` and throws an error if tags module is installed.

I think this is also the reason why tests are failing.

We should also maybe provide a migration script to uninstall tags module, if present, when this jcontent version is installed.

src/javascript/TagManager/TagManager.jsx
```diff
@@ -0,0 +1,350 @@
...
+                ['closeButton', 'closeAfter5s']
+            );
+            setEditNodeTarget(null);
+        } catch (error) {
+            notify(error.message, ['closeButton', 'noAutomaticClose']);
+        }
+    };
+
+    return (
+        <div className={styles.root} data-cm-role="tag-manager-root">
```
gflores-jahia: We should use `LayoutContent` moonstone template here to minimize boilerplate styling https://jahia.github.io/moonstone/?path=/docs/layouts-layoutcontent--docs

tagManager.cy.ts
gflores-jahia, line 1: Not needed.

tests/cypress/e2e/jcontent/tagManager.cy.ts
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    before(function () {
+        deleteSite(siteKey);
+        createSite(siteKey, {
+            templateSet: 'dx-base-demo-templates',
+            serverName: 'localhost',
+            locale: 'en'
+        });
+        enableModule('tags', siteKey);
```
gflores-jahia, line 61: We should remove this to test the requirement: _The tag manager will now appear for all sites, not just when tags are enabled on the site_

src/javascript/TagManager/register.js
```diff
@@ -0,0 +1,14 @@
...
+        targets: ['jcontent'],
+        label: 'jcontent:label.contentManager.navigation.manage.tags.title',
+        icon: <Tag/>,
+        isSelectable: true,
+        requiredPermission: 'tagManager',
```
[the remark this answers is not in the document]
gflores-jahia: This is not required

src/javascript/TagManager/TagManager.jsx
```diff
@@ -0,0 +1,350 @@
...
+
+    useEffect(() => {
+        setSearchTerm('');
+        setPage(0);
+        setSelectedTag(null);
+        setRenameTarget(null);
+        setDeleteTarget(null);
+        setEditNodeTarget(null);
+        setDrawerPage(0);
+    }, [siteKey]);
```
gflores-jahia: Manually resetting states when siteKey changes is a bit of an anti-pattern flag. Probably an alternative here is to put key={siteKey} on usage of <TagManager>, or create a wrapper component on TagManager with the key so React automatically remounts and resets these values.

https://react.dev/learn/you-might-not-need-an-effect

src/javascript/TagManager/TagManagerDialogs.jsx
```diff
@@ -0,0 +1,328 @@
...
+        name: PropTypes.string,
+        occurrences: PropTypes.number
+    }),
+    isOpen: PropTypes.bool,
+    isLoading: PropTypes.bool,
+    onClose: PropTypes.func.isRequired,
+    onConfirm: PropTypes.func.isRequired
+};
+
+RenameTagDialog.defaultProps = {
```
gflores-jahia: `defaultProps` will be a breaking change for React 19 so we should avoid using them on new code. We can use built-in javascript param default notation if needed. There are multiple instances of this in the code.

src/javascript/TagManager/TagManager.jsx
```diff
@@ -0,0 +1,350 @@
...
+        return compareStrings(left.name, right.name) * multiplier;
+    });
+};
+
+const getImpactedItemsCount = result => {
+    const counts = result?.workspaceResults?.map(workspaceResult => workspaceResult.processedCount) || [];
+    return counts.length > 0 ? Math.max(...counts) : 0;
+};
+
+export const TagManager = () => {
```
gflores-jahia, line 13: `TagManager` is handling a lot of responsibilities here. Consider maybe grouping some of related pieces together into its own component. 

For example the rename and delete actions and its related dialog components are used inside table component while edit node and delete node are all under the drawer component. it might make sense to push these down into their respective components.

A good rule of thumb would be to try to add them as close to the component level using them as possible.

The same could be argued for pagination states, which to me is more responsibility of the table, so it might makes sense for it to be handled there. 

Consider also maybe usage of custom hooks to bind these states together and manage them more efficiently. For example, it might make sense to group management of filter and sort mechanisms into a custom hook that exposes only what's needed/used in other modules. This makes it easier to separate concerns and understand the component as a whole.

src/javascript/TagManager/TagManagerDialogs.jsx
```diff
@@ -0,0 +1,328 @@
...
+    onSelect: PropTypes.func.isRequired
+};
+
+export const RenameTagDialog = ({siteKey, siteName, tag, isOpen, isLoading, onClose, onConfirm}) => {
+    const {t} = useTranslation('jcontent');
+    const [value, setValue] = useState(tag?.name || '');
+
+    useEffect(() => {
+        setValue(tag?.name || '');
+    }, [tag]);
```
gflores-jahia: Similar to https://github.com/Jahia/jcontent/pull/2265/changes#r3236954915 manual resetting of values is an antipattern. Consider usage of `key={renameTarget?.name}` to automatically remount and reset these values.

src/javascript/TagManager/TagManagerDialogs.jsx
```diff
@@ -0,0 +1,328 @@
+import React, {useEffect, useState} from 'react';
+import PropTypes from 'prop-types';
+import {Dialog, DialogActions, DialogContent, DialogContentText, DialogTitle} from '@material-ui/core';
+import {Button, Close, Input, Search, Typography} from '@jahia/moonstone';
+import {Trans, useTranslation} from 'react-i18next';
+import {useTagSuggestions} from './useTagSuggestions';
+import styles from './TagManager.scss';
+
+const TagSuggestions = ({label, suggestions, onSelect}) => {
```
gflores-jahia: Consider splitting the components/dialogs into its own files for better code org.

src/javascript/TagManager/TagManagerDrawer.jsx
```diff
@@ -0,0 +1,194 @@
...
+    onClose,
+    onPageChange,
+    onPageSizeChange,
+    onEditTagOnNode,
+    onDeleteTagOnNode
+}) => {
+    const {t} = useTranslation('jcontent');
+    const {language} = useSelector(state => ({
+        language: state.language
+    }), shallowEqual);
```
gflores-jahia: [minor] Converting to object and using shallowEqual is not needed here. Can be replaced with just `const language = useSelector(state => state.language);`

src/javascript/TagManager/TagManager.jsx
```diff
@@ -0,0 +1,350 @@
...
+            setPage(maxPage);
+        }
+    }, [page, pageSize, sortedTags.length]);
+
+    const refreshAfterMutation = useCallback(async () => {
+        await refetch();
+        await client.reFetchObservableQueries();
+        triggerRefetch(refetchTypes.CONTENT_DATA);
+        triggerRefetch(refetchTypes.CONTENT_TREE);
+    }, [client, refetch]);
```
gflores-jahia: I'm not clear why triggerRefretch is needed as we have our own tag data refetches already. as far as I can tell, tag manager doesn't depend on any of the external jcontent data which what triggerRefretch does. Also, client.retechObservableQueries() will refetch all active queries in the application, so maybe refetch could be enough here.

src/javascript/TagManager/TagManagerDialogs.jsx
```diff
@@ -0,0 +1,328 @@
...
+                            onClick={() => setValue('')}
+                        />
+                    ) : null}
+                    onChange={event => setValue(event.target.value)}
+                />
+                <TagSuggestions
+                    label={t('jcontent:label.contentManager.tagManager.rename.suggestions')}
+                    suggestions={suggestions}
+                    onSelect={setValue}
+                />
```
gflores-jahia: There's a bit of cohesion and repeated pattern with these components in both rename and edit tag dialogs. Consider encapsulating the input field, TagSuggestions, as well as value state and suggestions hook call as a separate component

src/javascript/TagManager/TagManagerDrawer.jsx
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        await onDeleteTagOnNode(deleteTarget.uuid);
+        setDeleteTarget(null);
+    };
+
+    return (
+        <>
+            {isOpen && (
+                <div className={styles.drawerLayer} data-cm-role="tag-manager-drawer-layer" onClick={onClose}>
```
gflores-jahia: **Update**: discussed with Felix. There are plans to have a common Drawer mechanism integrated but is not available yet https://github.com/Jahia/moonstone/issues/1363

So for now, it'd be good to at least split this into two components, one is a drawer container component to  handle UI styling and open/close behaviour aspects, and then another drawer content component for the drawer header and body. I see that the close button is currently within the drawer header so we should also separate these. This way, it'll be easier to migrate into the new design later on. 

Felix also recommends reuse of existing moonstone components with the display of the list of content in the drawer, but is unsure whether we should display this in table form or in card form (both we already have available as moonstone components). We keep this as-is for now.

----

There is currently an existing Drawer component but is based on the obsolete Material UI. I don't know if we should use this for consistency or if there are any design plans for revamping or unifying the Drawer component in moonstone, or if this will be an ok stand-in for now. Maybe need to discuss with @ffffffelix .

The implem also differs a bit from the design so we'll also need to confirm with Felix to see if this ok for now or if will need to be revised.

src/javascript/TagManager/TagManager.jsx
```diff
@@ -0,0 +1,350 @@
...
+                    siteKey,
+                    tag: renameTarget.name,
+                    newName
+                }
+            });
+            const result = mutationData?.admin?.jahia?.tagManager?.renameTag;
+            await refreshAfterMutation();
+            if (selectedTag === renameTarget.name) {
+                setSelectedTag(newName);
+            }
```
gflores-jahia: possible race condition for setting selected tags after the refresh. Consider an order swap.

src/javascript/TagManager/TagManager.jsx
```diff
@@ -0,0 +1,350 @@
...
+        try {
+            const {data: mutationData} = await renameTagOnNode({
+                variables: {
+                    siteKey,
+                    tag: selectedTag,
+                    newName: newTag,
+                    nodeId: editNodeTarget.uuid
+                }
+            });
+            const result = mutationData?.admin?.jahia?.tagManager?.renameTagOnNode;
```
gflores-jahia: I noticed workspaceResults has failedCount, failedPaths as part of possible mutation result values. It could be a good idea to include in the results and display an error notif if they are non-zero/non-null.

This can also apply for the delete, delete node actions.

gflores-jahia (changes requested):
Overall a good initial implementation and covers all requirements. Also appreciate added cypress tests as well. I do have general comment around code organization and using React best practices, specifically around TagManager component having too much responsibilities. We should practice single responsibility principle and keeping related states, components grouped as much as possible and closer to the component level that is being used.

Also something for consideration, but we can try to implement this as a separate package in the monorepo design we are planning on putting in place for jcontent, but since this will be the first full UI extension package, I don't know if we anticipate any tech roadblocks issues considering jcontent release timeline. @GauBen I don't know if you think we should attempt this now as part of the scope of this bounty or hold off on it later.

GauBen:
+2082 -0? This is a fantastic candidate for isolation @gflores-jahia!

src/javascript/TagManager/TagManager.jsx
smonier: TagManagerRoute wrapper in register.js reads siteKey from Redux and passes it as key to TagManager — React handles the remount and state reset automatically. The manual useEffect reset is removed from TagManager.jsx.

src/javascript/TagManager/TagManager.scss
```diff
@@ -0,0 +1,331 @@
...
+}
+
+.selectedRow {
+    background-color: #00A0E3;
+}
+
+.selectedRow :global(.moonstone-typography),
+.selectedRow :global(.moonstone-button:not(:hover):not(:focus-visible)) {
+    color: var(--color-white);
+}
```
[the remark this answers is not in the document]
smonier: replace hardcoded color with Moonstone design token in TagManager
background-color: var(--color-accent);

src/javascript/TagManager/register.js
smonier: rename to     registry.add('adminRoute', 'jctagsmanager', {...}

src/javascript/TagManager/TagManager.jsx
smonier: Added LayoutContent to the Moonstone import, replaced the manual root + pageHeader + content div stack with <LayoutContent hasPadding
   header={...} content={...}/>, and moved the three dialogs into a fragment alongside it.                                                               
  - TagManager.scss: Deleted .root, .pageHeader, .pageHeaderInner, and .content — all now handled by LayoutContent — and removed the dead .content rule  
  from the media query.

src/javascript/TagManager/TagManagerDialogs.jsx
smonier: All defaultProps across TagManagerDialogs.jsx, TagManagerDrawer.jsx, and TagManagerTable.jsx have been replaced with inline JS default parameter values and pushed.

src/javascript/TagManager/TagManager.jsx
smonier: - TagManager.utils.js (new) — shared getImpactedItemsCount helper                                                                                      
  - useTagFilter.js (new) — encapsulates searchTerm, sort, filtered/sorted tag computation                                                               
  - TagManager.jsx — now just: data fetching, selectedTag, refreshAfterMutation, render                                                                  
  - TagManagerTable.jsx — now owns: search UI, loading/error/empty states, pagination, rename/delete mutations and dialogs                               
  - TagManagerDrawer.jsx — now owns: drawer pagination, deleteNode/editNode mutations and dialogs (EditNodeTagDialog moved here from the parent)

src/javascript/TagManager/TagManagerDialogs.jsx
smonier: Removed the useEffect reset pattern from both RenameTagDialog and EditNodeTagDialog, dropped the now-unused useEffect import, and added          
  key={renameTarget?.name} / key={editNodeTarget?.uuid} at the call sites so React handles the reset automatically on remount.

src/javascript/TagManager/TagManagerDialogs.jsx
smonier: TagManagerDialogs.jsx is gone, replaced by five focused files:                                                                                   
                                                                                          
  - TagSuggestions.jsx — shared sub-component                                                                                                            
  - RenameTagDialog.jsx                                                                                                                                  
  - DeleteTagDialog.jsx                                                                                                                                  
  - DeleteNodeTagDialog.jsx                                                                                                                              
  - EditNodeTagDialog.jsx                                                                                                                              
                                                                                                                                                         
  TagManagerTable and TagManagerDrawer now import directly from each dialog's own file.

src/javascript/TagManager/TagManager.jsx
smonier: Removed useApolloClient, useCallback, triggerRefetch, and the two extra refetch calls — refetch from useQuery is now passed directly as onMutationComplete.

src/javascript/TagManager/TagManagerDialogs.jsx
smonier: TagNameInput.jsx now owns the value state, useTagSuggestions call, Input rendering, and TagSuggestions — and notifies the parent via onChange so the dialog can still compute normalizedValue for the confirm button's disabled check. Both RenameTagDialog and EditNodeTagDialog are significantly slimmer.

src/javascript/TagManager/TagManager.jsx
smonier: onTagRenamed now fires before onMutationComplete(), so selectedTag is updated to the new name before the refetch triggers re-renders — the drawer will query with the correct tag name from the start.

src/javascript/TagManager/TagManager.jsx
smonier: Changes across 7 files:                                                                                                               
                                                                                          
  - GQL queries — all four mutations now fetch failedCount and failedPaths per workspace                                                                 
  - TagManager.utils.js — added getFailedCount(result) alongside getImpactedItemsCount                                                                   
  - TagManagerTable.jsx — rename and delete handlers now show a persistent error notification if failedCount > 0                                         
  - TagManagerDrawer.jsx — deleteTagOnNode and editTagOnNode handlers do the same                                                                        
  - en/fr/de.json — new partialFailure translation key added to all three locales

src/javascript/TagManager/register.js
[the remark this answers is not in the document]
smonier: no longer present

src/javascript/TagManager/TagManagerDrawer.jsx
smonier: waiting on the common drawer, nothing done, kept the existing one for now.

tagManager.cy.ts
gflores-jahia: This one still exists.

gflores-jahia (commented):
Hey @smonier thanks a lot for all the changes, it looks a lot more readable. I think we're getting close, just need to clean up minor stuff. I added quick follow-up comment.

There's also still some static analysis issues that needs to be looked at https://github.com/Jahia/jcontent/actions/runs/26037511225/job/76539657445?pr=2265

And finally tag manager tests are still failing, and can reproduce locally. From what I've seen it's mostly because it navigates to the old tag manager and not the new one.

src/javascript/TagManager/register.js
gflores-jahia: > We should also maybe provide a migration script to uninstall tags module, if present, when this jcontent version is installed.

Checking on the channel if we want to add above https://jahia.slack.com/archives/C0APR9GDPNX/p1779829539295259

gflores-jahia: From Romain, ticket to be created for the migration script.

src/javascript/TagManager/TagManagerDrawer.jsx
smonier: Done. Simplified the selector and removed the now-unused shallowEqual import.

src/javascript/TagManager/TagManagerDrawer.jsx
smonier: Moonstone 2.20 drawer has been implemented instead of the original custom one

smonier:
Went through each items from @gflores-jahia review (except the migration), implemented the new drawer from moonstone 2.20, updated Cypress test, fixed all Lint errors.

gflores-jahia (commented):
Hi, everything looks good now, except for a last-min request. Since this PR depends on some graphql api changes, we need to bump our dependency version to sync up with the to-be-released version https://github.com/Jahia/jcontent/blob/main/pom.xml#L52

Also heads up that we shouldn't merge this PR until gql changes have been merged and available as a snapshot.

tests/cypress/e2e/jcontent/tagManager.cy.ts
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+    beforeEach(function () {
+        deleteNode(contentRootPath);
+        createTestContent();
+        cy.loginAndStoreSession();
+    });
+
+    it('supports viewing usages for special-character tags', () => {
+        const tagManager = TagManager.openFromAdditionalApps(siteKey, 'en');
```
gflores-jahia, line 76: Failing test. based on the test failure recording, I think this one is missing loading of the tag manager page (to be confirmed)

gflores-jahia (commented):
Thank you for the dependency change @smo. Code looks ok to me now. But there are also still some Sonarqube analysis failures that still needs to be looked at and addressed.

For tests, it looks like tag manager suite is still failing. See one of my code comments below, but also I think graphql-core PR still needs to be merged to get these to pass. 

Also there is still that empty test file that still needs to be removed. https://github.com/Jahia/jcontent/pull/2265/changes#r3306562839

btw, I didn't realize you'd already made the dep change since last week. For future PR changes, it would be helpful to click the refresh button to Re-request review after addressing feedback. It helps me get notified that it's ready to be looked at. Thanks!

tests/cypress/e2e/jcontent/tagManager.cy.ts
smonier: Fixed. Added data-cm-role="tag-manager-loader" to the loading container in TagManagerTable and updated openFromAdditionalApps in the page object to wait for that loader to disappear before returning — ensuring the initial GraphQL fetch completes before the test interacts with rows.

sdusolle (approved):

tests/provisioning-manifest-build.yml
```diff
@@ -3,5 +3,10 @@
 
 - include: 'provisioning.yml'
 
+- installBundle:
+    - 'mvn:org.jahia.modules/graphql-dxm-provider'
+  autoStart: true
+  uninstallPreviousVersion: true
+
```
gflores-jahia, line 10: This shouldn't be needed as snapshot should already install 3.6.1-SN

gflores-jahia (changes requested):
@smo there's been a reorg of the cypress tests, and looks like tagmanager suite didn't execute on the last run - `tagManager.cy.ts` needs to now be in one of `/shard1` or `/shard2` folder to be included. There's also this minor clean-up still to be done btw https://github.com/Jahia/jcontent/pull/2265/changes#r3306562839

I also still think https://github.com/Jahia/graphql-core/pull/597 needs to be merged first to have the graphql API available to get these tests to pass.

rknj:
Just merged the GraphQL PR and triggered a new run for this PR: https://github.com/Jahia/jcontent/actions/runs/28954268008

pom.xml
```diff
@@ -49,7 +49,7 @@
         <yarn.arguments>build:production</yarn.arguments>
         <jahia-module-signature>MC0CFHaCJ6hx88BuPxemDd5E4lGxj79hAhUAlhPBLrK5Zc3J0O/yrj1SeY5YKuM=</jahia-module-signature>
         <require-capability>osgi.extender;filter:="(osgi.extender=org.jahia.bundles.blueprint.extender.config)"</require-capability>
-        <jahia-depends>jahia-ui-root=1.10.0,app-shell=2.7,graphql-dxm-provider=3.5.0</jahia-depends>
+        <jahia-depends>jahia-ui-root=1.10.0,app-shell=2.7,graphql-dxm-provider=3.6.1</jahia-depends>
```
gflores-jahia, line 52: Since it's been merged after 3.7.0 release, we need to adjust this now to new version 3.8. This PR https://github.com/Jahia/jahia-pack-private/pull/1028 also needs to be merged prior to re-running the test.

rknj: @TheKlem @romain-pm are you OK with this dependency? Cloud customer that would like to use recent jContent version on a previous Jahia version will ask for a GQL upgrade

rknj: Otherwise @smonier @gflores-jahia is there a way to display the tag manager on specific conditions?

smonier: We moved from webflow to GraphQL for the new tag manager which requires now graphQL 3.6.1, we may want to test the version of GQL before rendering it ? the extension has been renamed to 'jctagsmanager' so shouldn't conflict with the old one. I'll let @gflores-jahia comment, he is the jContent master.

gflores-jahia: > @smonier @gflores-jahia is there a way to display the tag manager on specific conditions?

~~Not currently, but I created a PoC and we can add a new `requiredBundle` option when registering for admin route. It would be something like:~~

~~registry.add('adminRoute', 'jctagsmanager', {
        [...]
        requiredBundle: {bundle: 'graphql-dxm-provider', versionRange: '>=3.8'}
    });~~

~~We can add this in also as part of that PR, so no need to add it here, but we can remove the hard requirement in the pom.xml~~

Actually the above doesn't quite work as the current gql api `admin.tools.bundles` doesn't have a filter for active bundles. It can be that two versions of the bundle is installed but no way to filter which one is currently active. we'll need to improve this in the _tools_ module if we need this filter functionality.

The only other workaround is to specifically expose the current gql-dxm-provider version as part of jcontent config e.g. contextJsParameters.config.graphqlDxmProviderVersion, then add the check in register.js for tag manager. I created [this](https://github.com/Jahia/jcontent/pull/2545) as follow-up PR after merge.

As for migration from old tags manager, we can add a migration script in graphql to uninstall old tags module when installed.

rknj: @smonier @gflores-jahia I've discussed this with @romain-pm and we'll not push this in the upcoming 3.7.0. So please wait for the 3.7.0 to be released before pushing this PR.

smonier:
@jkevan @gflores-jahia 

Deleting the old tags module from Jahia breaks Jahia System Templates  (templates-system) that has a dependency on it. Should I address it as it is part of the bounty ? My mistake that i didn't see it before.

gflores-jahia:
@smonier I think from your analysis, it's being depended on by a few more modules https://github.com/Jahia/jcontent/blob/retire-tags-module-analysis/docs/tags-module-retirement-analysis.md#what-retiring-tags-would-actually-involve.

I'm not sure if these migration is also in scope of this bounty or not. at the minimum if we can go with your alternative option of just unregistering the UI admin route, but only when the new tag manager route is activated then I think it's a minimum viable option and we can tackle nodetype migration independently (whether through bounty or not).

gflores-jahia:
also it looks like tests are failing after the sync due to unresolved advancedOptions page object: 

```
Error: Webpack Compilation Error — Module not found: Error: Can't resolve './advancedOptions' in '/home/jahians/cypress/page-object'.
```
It was removed [here](https://github.com/Jahia/jcontent/pull/2537) because we do not carry advanced options anymore.

smonier:
Follow-up on the remaining review threads (pushed as `fa8d8f38` + `04f04399`):

- **All 9 integration suites red** — root cause was mine: `tests/cypress/page-object/index.ts` still exported `AdvancedOptions` from a file that no longer exists on `main` (kept by mistake when rebasing). Every spec imports that index, so all suites failed at webpack compile with 0 tests run. Fixed in `fa8d8f38`.
- **`register.js` — conflict with the legacy `tagsmanager` route** — our route is `jctagsmanager`, and in addition jContent now removes the legacy `tagsmanager` admin route via a `jahiaApp-init:9999` callback (same pattern `remotepublish` uses), *only* when the new Tag Manager was actually registered (provider ≥ 3.8). The tags module itself, its node types, actions and components are untouched — the migration/retirement question stays with the ticket Romain asked for (analysis in #2768).
- **`tagManager.cy.ts` — `enableModule('tags', siteKey)`** — removed, so the test now checks the requirement that the Tag Manager appears regardless of the tags module being enabled on the site.
- **`provisioning-manifest-build.yml` — explicit `installBundle` of graphql-dxm-provider** — removed; `jahia-pack-private#1028` is merged so the `8-SNAPSHOT` image already ships ≥ 3.8.
- **`pom.xml` `jahia-depends`** — superseded by the runtime version gate from #2545 (`jahia-depends=…,graphql-dxm-provider=3.5`, gate `>=3.8`).
- **`tagManager.cy.ts:76` (failing test)** — the page object already waits for the loader to disappear; couldn't be assessed while the suites didn't compile, so let's see this CI run.

smonier:
Update after `b8546e3a`: the Tag Manager spec was not being run by CI at all — since the suites were split into shards (#2465), the configs only match `cypress/e2e/jcontent/shard{1,2}/**`, and `tagManager.cy.ts` sat one level up. Moved it into `shard1`.

First real run since then, on this head (`cypress.config-jcontent.ts`):

```
jcontent/shard1/tagManager.cy.ts
  ✓ supports viewing usages for special-character tags
  ✓ renames a tag on a single content item from the side panel
  ✓ renames and deletes a site tag from the table
  ✓ removes a tag from one content item with confirmation from the side panel
  5 passing, 0 failing
```

That covers the `tagManager.cy.ts:76` thread. All 9 integration suites, build, unit tests, lint and Sonar are green on the PR.

jahia-carp:
This pull request also unblocks a test migration in [Jahia/selenium](https://github.com/Jahia/selenium). The Tag manager suite it adds covers most of a QA scenario that still runs in Selenium today.

[selenium#1492](https://github.com/Jahia/selenium/issues/1492) tracks the move of `testTags` from `CECrossFeaturesTest` to Cypress. The QA scenario has four steps:

- Add tags on two contents in the Content Editor.
- Open jContent > Additional > Tag manager and list the tags.
- Check the occurrence count of a tag, then delete that tag.
- Reopen the content and check that the tag is gone.

`tests/cypress/e2e/jcontent/shard1/tagManager.cy.ts` in this pull request covers the Tag manager part of that scenario. When this pull request merges, we can delete the Selenium class and close [selenium#1491](https://github.com/Jahia/selenium/issues/1491) and [selenium#1492](https://github.com/Jahia/selenium/issues/1492).

One step of the QA scenario has no coverage at the UI level. After a site tag is deleted from the table, the scenario reopens the content and checks the Tags field of the Content Editor. [graphql-core](https://github.com/Jahia/graphql-core) covers the data side in `tests/cypress/e2e/api/tagManager.cy.ts`, with the test "removes the tag from all nodes in both workspaces". The gap is therefore small, and it is not a defect of this pull request.

Here is a sketch of that assertion, for `tagManager.cy.ts`:

```ts
it('removes a deleted site tag from the content editor field', () => {
    const tagManager = visitTagManager();
    tagManager.search(commonTag).openDelete(commonTag).confirmDelete();

    const jcontent = JContent.visit(siteKey, 'en', 'content-folders/contents/tagmanager-e2e');
    const contentEditor = jcontent.editComponentByRowName('Tag Alpha');
    contentEditor.switchToAdvancedMode();
    contentEditor.openSection('Classification and Metadata');

    const tagField = contentEditor.getField(TagField, 'jmix:tagged_j:tagList');
    tagField.getTags().should('have.length', 1);
    tagField.assertTagText(alphaTag, 0);
    contentEditor.cancelAndDiscard();
});
```

Please do not hold the merge for this. We can open a follow-up pull request after the merge, and we are happy to write it ourselves.

gflores-jahia (approved):

gflores-jahia:
Lgtm, thank you for your patience :)

PR #2275, opened 2026-04-10
fix: add missing i18n label for j:bindedComponent in Content Editor

pvollenweider:
## Summary

- Added `labelKey` to the `j:bindedComponent` field in `jmix_bindedComponent.json` — without it, the field fell back to its technical name (`j_bindedComponent`) in the UI
- Added translations for the new label in EN, FR, and DE

## Changes

- `forms/jmix_bindedComponent.json`: added `"labelKey": "label.contentEditor.boundComponent.bindedComponent.label"`
- `jcontent_en.properties`: `Component link`
- `jcontent_fr.properties`: `Lien vers le composant`
- `jcontent_de.properties`: `Komponentenverknüpfung`

Fixes #4933


src/main/resources/resources/jcontent_de.properties
```diff
@@ -1,20 +1,21 @@
 label.jContent=jContent
 label.permission.jContentAccordions.description=Berechtigung zum Zugriff auf jContent. Stellen Sie sicher, dass dieser Rolle auch mindestens eine Accordion-Berechtigung erteilt wird.
-label.permission.additionalAccordionAccess.description=Zeigen Sie das "Zus�tzliche" Akkordeon in jContent an. Beachten Sie, dass f�r jeden Eintrag im Akkordeon m�glicherweise zus�tzliche Berechtigungen gelten.
+label.permission.additionalAccordionAccess.description=Zeigen Sie das "Zus\u00E4tzliche" Akkordeon in jContent an. Beachten Sie, dass f\u00FCr jeden Eintrag im Akkordeon m\u00F6glicherweise zus\u00E4tzliche Berechtigungen gelten.
```
gflores-jahia, line 3: Looks like some encoding issues with the props file and causing some unintended edits. I know editing at least on intellij should work/auto-identify file encoding.

Current encoding is ISO-8859-1 if it helps.

pvollenweider: The encoding changes in the properties files are the result of IntelliJ's native2ascii conversion, which replaces raw non-ASCII characters with their \uXXXX Unicode escape equivalents. This is a standard and recommended practice for Java .properties files, as specified in the https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html 

IDEs like IntelliJ auto-detect the encoding and display the escaped sequences as readable characters, so there is no impact on readability during development. 

encoding

gflores-jahia: Thanks for the explanation! You're right that \uXXXX escapes are technically valid and see no issues with the translations in the UI. But that being said, from the given doc link:

> [...] input/output stream is encoded in ISO 8859-1 character encoding. Characters that cannot be directly represented in this encoding can be written using Unicode escapes 

Our .properties files are already required to be ISO-8859-1 encoded, and \u escapes should be reserved for characters outside the ISO-8859-1 range. French characters like é, à, ê are already within ISO-8859-1, so I believe they don't really need escaping. We have these style of formatting already in place for other projects like https://github.com/Jahia/jahia-private/blob/master/war/src/main/resources/JahiaInternalResources_fr.properties and https://github.com/Jahia/jahia-private/blob/master/war/src/main/resources/JahiaInternalResources_de.properties. Given a choice, I also personally would prefer to keep the non-escaped characters for readability outside of intellij

For now, I would much rather keep the diffs to a minimum and not deviate from what we have in our existing repos so far. But I can also check with rest of the team to see if we need to do these conversions in our repos to keep them consistent.

gflores-jahia (approved):

PR #2279, opened 2026-04-15
fix: Handle template redirection correctly

### LINKED ISSUE #2203: Page Builder fails to load pages with dot (.) in system name – redirects back to current page instead

- Jahia 8.2.0.x
- Page Builder (JContent UI)
- Custom site with pages having system names that include a dot (e.g., `balbala.123`)
- Issue reproducible in local environment using:
  - Default modules only
  - Node path /sites/mySite/home/balbala.123

### Steps to reproduce

1. Create a new content page with a system name that contains a dot (`.`), e.g. balbala.123
2. From Page Builder, navigate to /jahia/jcontent/mySite/en/pages/home
3. Add a link or button pointing to the page /sites/mySite/home/balbala.123
4. Click the link while in Page Builder

### Current Behavior

Instead of navigating to /cms/editframe/default/en/sites/mySite/home/balbala.123.html, the Page Builder view reloads the current page (home) and appends a query string like:
```
/jahia/jcontent/mySite/en/pages/home?params=(sub:!f)
```
The new page is not loaded, and the user stays on the same page.
This seems to be related to a logic error in EditFrame.jsx, particularly the useEffect block that generates the encodedPath:
```
const encodedPath = path.replace(/[^/]/g, encodeURIComponent) + (template === '' ? '' : `.${template}`);
```
This line incorrectly encodes the dot `.` in the page name, resulting in a mismatch when comparing path === framePath, which then fails the logic intended to update the iframe.

Example scenario:
- Page URL: http://127.0.0.1:8080/jahia/jcontent/mySite/en/pages/home
- Intended link: /cms/editframe/default/en/sites/mySite/home/balbala.123.html
- Observed redirect: http://127.0.0.1:8080/jahia/jcontent/mySite/en/pages/home?params=(sub:!f)

### Desired Behavior

When clicking on a link to a page with a dot in its system name, the iframe should correctly navigate to the target page, such as:
```
/cms/editframe/default/en/sites/mySite/home/balbala.123.html
```
The encoding and comparison logic should handle system names containing `.` without triggering a fallback to the current page.
Navigation in Page Builder should behave the same way regardless of whether a page’s system name contains dots or not.
The issue seems to be in content/src/javascript/JContent/EditFrame/EditFrame.jsx around line 206...

AKarmanov:
### Description
Do not resolve template from path. This changes the work done here: https://github.com/Jahia/jira-archives/issues/15703 and here https://github.com/Jahia/jcontent/pull/1739. Template resolves correctly now and is passed via params.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/.chachalog/VGkEM-Go.md
+++ b/.chachalog/VGkEM-Go.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": minor
+---
+
+Handle template redirection corectly (#2279)
--- a/.chachalog/VGkEM-Go.md
+++ b/.chachalog/VGkEM-Go.md
@@ -3,4 +3,4 @@
 "@jahia/jcontent": minor
 ---
 
-Handle template redirection corectly (#2279)
+Fix redirection issue in page builder when system name contains . (#2279)
--- a/src/javascript/JContent/redux/JContent.redux.js
+++ b/src/javascript/JContent/redux/JContent.redux.js
@@ -26,10 +26,6 @@ const apps = {
             const accordion = registry.get('accordionItem', mode);
             const viewMode = localStorage.getItem('jcontent-previous-tableView-viewMode-' + site + '-' + mode) || accordion?.tableConfig?.defaultViewMode || 'flatList';
             let template = '';
-            if (viewMode === JContentConstants.tableView.viewMode.PAGE_BUILDER && path.lastIndexOf('.') > 0) {
-                template = path.substring(path.lastIndexOf('.') + 1);
-                path = path.substring(0, path.lastIndexOf('.'));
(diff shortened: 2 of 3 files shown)


.chachalog/VGkEM-Go.md
```diff
@@ -0,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": minor
+---
+
+Handle template redirection corectly (#2279)
```
gflores-jahia: Maybe tie the changelog a bit to the original issue e.g. `fix redirection issue in page builder when system name contains .`

gflores-jahia (dismissed):

gflores-jahia (approved):

PR #2281, opened 2026-04-16
fix: keep dynamic fieldsets in editForm response to restore addMixin mechanism

pvollenweider:
## Summary

- Empty `jmix:templateMixin` fieldsets (`hasEnableSwitch=false`) were filtered out of the GraphQL response by the visibility logic introduced in BACKLOG-21449 (commit `1871f272`, PR #999).
- As a result, the frontend's `getDynamicFieldSets()` had no knowledge of these fieldsets, and `getMixinsToMutate()` never included them at save time — silently breaking the `addMixin` choicelist mechanism.
- This regression only affects mixins that extend `jmix:templateMixin` and have no visible properties (all hidden or none declared).

## Root cause

`EditorFormServiceImpl` line 220 (before fix):
```java
fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isHasEnableSwitch() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
```
For a `jmix:templateMixin` mixin with no visible fields: `hasEnableSwitch=false` + `fields.isEmpty()=true` → `visible=false` → filtered out → frontend never sees it.

## Fix

Replace `isHasEnableSwitch()` with `isDynamic()` in the visibility condition:
```java
fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isDynamic() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
```

A dynamic fieldset must always be included in the response — even when empty — because it may be activated programmatically via `addMixin`. The frontend's display filter (`filterFieldSets`) already handles not rendering inactive dynamic fieldsets, so there is no visual regression.

Non-dynamic empty fieldsets continue to be filtered out as before.

## Verification

- Worked correctly on Jahia 8.1.6.2 (predates [BACKLOG-21449](https://github.com/Jahia/jcontent/pull/999/changes/86824acb191f824c33c0d5589307d644d09e6b68))
- Broken on all versions after commit `1871f272` when using `jmix:templateMixin` mixins with no visible fields and the `addMixin` choicelist mechanism

Closes https://github.com/Jahia/jahia-private/issues/4940

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
@@ -217,7 +217,7 @@ public class EditorFormServiceImpl implements EditorFormService {
                         .filter(Field::isVisible)
                         .collect(Collectors.toList()));
 
-                    fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isHasEnableSwitch() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
+                    fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isDynamic() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
                 }
 
                 // Remove empty fieldSets - only keep empty dynamic field set which do not have matching mixin in another section
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
@@ -217,14 +217,21 @@ public class EditorFormServiceImpl implements EditorFormService {
                         .filter(Field::isVisible)
                         .collect(Collectors.toList()));
 
-                    fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isDynamic() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
+                    fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isHasEnableSwitch() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
                 }
 
-                // Remove empty fieldSets - only keep empty dynamic field set which do not have matching mixin in another section
-                // (if the dynamic mixin is present in multiple sections, keep only the non-empty ones)
+                /*
+                 * Keep fieldSets that passes one of the following criteria:
+                 *  - Always present
+                 *  - Dynamic and unique to this section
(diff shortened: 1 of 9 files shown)


src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
```diff
@@ -217,7 +217,7 @@ private Form getEditorForm(ExtendedNodeType primaryNodeType, JCRNodeWrapper exis
                         .filter(Field::isVisible)
                         .collect(Collectors.toList()));
 
-                    fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isHasEnableSwitch() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
+                    fieldSet.setVisible(fieldSet.isVisible() && (fieldSet.isDynamic() || fieldSet.getFields().stream().anyMatch(Field::isVisible)));
```
gflores-jahia: @pvollenweider You're right that dynamic fields still needs to be included in the response. But the change might not correspond semantically to the visibility flag (e.g. not all dynamic mixins are visible).

As an alternative solution, maybe the filtering of empty fieldset should be done on [this section](https://github.com/Jahia/jcontent/pull/2281/changes#diff-db640e6e30e0ce63d7877375677f817d5f9fce6c6b22342665cfe63036a5c2d7R227) instead where we remove empty fieldSets and don't immediately filter non-visible mixins:

```
/*
 * Keep fieldSets that passes one of the following criteria:
 *  - Always present
 *  - Dynamic and unique to this section
 *    - Dynamic single-occurrence fieldsets are always kept — this includes jmix:templateMixin fieldsets
 *      with no visible fields that are needed for addMixin tracking even though visible=false
 *  - Visible and non-empty.
 */
section.setFieldSets(section.getFieldSets().stream()
    .filter(fs -> (fs.isAlwaysPresent() != null && fs.isAlwaysPresent())
        || (fs.isDynamic() && fieldSetsMap.get(fs.getName()).size() == 1)
        || (fs.isVisible() && !fs.getFields().isEmpty()))
    .collect(Collectors.toList()));
```

Lmk what you think with this fix. I also added a cypress tests for this PR.

gflores-jahia (commented):
Thank you for the fix ! Verified that it is a regression when using jmix:templateMixin and addMixin. I proposed and updated an alt solution below. Will confirm with Philippe about this change.

gflores-jahia:
Fixed some issues around around our tests. 

I noticed we were starting up browsing nodes even though cluster is not enabled. 

I also noticed that when crashes happen that video captures are not fully cleaned so I make sure that the process is completed before copying it over to our artifacts folder.

AKarmanov (approved):

PR #2288, opened 2026-04-22
fix: Update sanitizer policy to allow links in tooltips

### LINKED ISSUE #2054: Provide more formating possibilities in UI tooltips
Some customers use heavily the UI tooltips to guide their editors but 2 shortcomings has been raised by them
- Impossible to put links in UI tooltips
- UI tooltips are not visible enough (grey on white)

### Story
As a developer
I want to add links in tooltips

## Testable scenarios

**Setup**
Given

**Case 1**
When I insert a < a href> instruction in a ui tooltip
Then the tlink appear in Content Editor and is clickable

**Case n**
If I insert any other htlml tag, this one is ignored when rendered in the GUI and doesn't even appear in the code.

### Test Strategy
Cypress tests?

### Tech Notes
What is the best strategy to allow/forbid tags ? Knowing that those tvalues are in resource bundles, in seems hard to do a validation when RB are saved because html can be authorized somewhere else in the app and in a RB. Therefore it seems that it's more at runtime that we should "clean" the html, but this point has to be discussed.

⚠️ Security: do we need to sanitize the href data?

### Labels 
En: 
Fr: 
De:

AKarmanov:
### Description
Update sanitizer policy to allow links in tooltips. 

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/Sanitizer.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/Sanitizer.java
@@ -1,0 +1,24 @@
+package org.jahia.modules.contenteditor.api.forms;
+
+import org.owasp.html.HtmlPolicyBuilder;
+import org.owasp.html.PolicyFactory;
+import org.owasp.html.Sanitizers;
+
+/**
+ * This class provides custom sanitizers for various form fields.
+ */
+public class Sanitizer {
+
+    public static final PolicyFactory DESCRIPTION_FORMATTING = new HtmlPolicyBuilder()
+        .allowCommonInlineFormattingElements()
+        .allowStandardUrlProtocols()
+        .allowElements("a")
+        .allowAttributes("href").onElements("a")
+        .toFactory();
+
+    public static final PolicyFactory COMMON_FORMATTING = Sanitizers.FORMATTING;
+
+    private Sanitizer() {
+        /* This utility class should not be instantiated */
+    }
+}
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/Sanitizer.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/Sanitizer.java
@@ -11,7 +11,7 @@ public class Sanitizer {
 
     public static final PolicyFactory DESCRIPTION_FORMATTING = new HtmlPolicyBuilder()
         .allowCommonInlineFormattingElements()
-        .allowStandardUrlProtocols()
+        .allowUrlProtocols("http", "https")
         .allowElements("a")
         .allowAttributes("href").onElements("a")
         .toFactory();
--- a/.chachalog/OtSkZEwG.md
+++ b/.chachalog/OtSkZEwG.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": minor
+---
+
+Update sanitizer policy to allow links in tooltips (#2288)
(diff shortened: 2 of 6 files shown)


src/main/java/org/jahia/modules/contenteditor/api/forms/Sanitizer.java
```diff
@@ -0,0 +1,24 @@
...
+ * This class provides custom sanitizers for various form fields.
+ */
+public class Sanitizer {
+
+    public static final PolicyFactory DESCRIPTION_FORMATTING = new HtmlPolicyBuilder()
+        .allowCommonInlineFormattingElements()
+        .allowStandardUrlProtocols()
+        .allowElements("a")
+        .allowAttributes("href").onElements("a")
+        .toFactory();
```
gflores-jahia, line 17: @bpapez need to get your opinion on this one as we're trying to allow links in UI tooltips. Plugging this in AI gives me this and don't know if you agree on some/all of them. (For me, at least whitelisting the protocols could be good to add)

Note that this will be something specified in cnd definitions (I don't know if there are other ways for this to be defined) so less exposed to user inputs so maybe leaving as-is is ok too.

> What’s missing / risky
> 1. No rel="noopener noreferrer"
> 
> If users can add links, you must prevent tabnabbing when links open in a new tab.
> 
> Right now, nothing enforces this.
> 
> 👉 Fix:
> 
> .requireRelNofollowOnLinks()
> 
> Optionally also enforce:
> 
> .allowAttributes("target").onElements("a")
> 
> …but only if you control it (see below).
> 
> 2. target="_blank" handling (important)
> 
> If you allow target, users can inject _blank and exploit window.opener.
> 
> You either:
> 
> Don’t allow target at all (simplest + safest), OR
> Allow it but enforce rel protections.
> 
> Best practice:
> 
> .allowAttributes("target").matching(true, "_blank").onElements("a")
> .requireRelNofollowOnLinks()
> 3. No URL attribute constraints
> 
> Right now, any valid protocol URL is allowed. That includes:
> 
> mailto:
> ftp:
> 
> That might be fine—but if you want stricter control, restrict to only http and https.
> 
> Example:
> 
> .allowUrlProtocols("http", "https")
> 4. Missing title (optional UX)
> 
> Not a security issue, but often expected:
> 
> .allowAttributes("title").onElements("a")
> 5. No link text enforcement
> 
> This isn’t something the sanitizer handles, but remember:
> 
> <a href="..."></a> (empty links) are allowed
> Could be abused for UI tricks
> 
> Usually handled at validation layer, not sanitizer.
>
> Recommended hardened version:
> ```
>public static final PolicyFactory DESCRIPTION_FORMATTING =
>    new HtmlPolicyBuilder()
>        .allowCommonInlineFormattingElements()
>        .allowElements("a")
>        .allowAttributes("href").onElements("a")
>        .allowAttributes("title").onElements("a")
>        .allowUrlProtocols("http", "https")
>        .requireRelNofollowOnLinks()
>        .toFactory();
> ```

bpapez: Regarding the first 2 points and the tabnabbing protection, I once had a discussion during a SECURITY ticket analysis with Baptiste and we found that the OWASP HTML Sanitzier is automatically adding `rel="noopener noreferrer"`  for links with` target="_blank"` , even if nothing is specified in the policy. One would need to add `skipRelsOnLinks(noopener, noreferrer)` in the policy in order to not get it. Also I found that all modern browsers are automatically protecting against tabnabbing.

So for me the only recommended change from security perspective is the exchange of  `.allowStandardUrlProtocols() ` with `.allowUrlProtocols("http", "https")`.

gflores-jahia (approved):

PR #2289, opened 2026-04-22
fix: removed seo permission from visibility tab and hardeded components for nulls

### LINKED ISSUE #2247: visibility tab has a dependency to seotab permission

8.2.2.1

### Steps to reproduce

Install env with digital
login with "mathias" and check on a content the visibilty tab (works fine)
login with root, modify the editor role, and remove permission "View seo tab"
login again with "mathias" and check again the visibilty tab, you run in an error:

### Current Behavior

Image

### Desired Behavior

The visibility tab has its own permission, "View visibility tab". So the "View seo tab" permission shouldn't have an effect on the visibility tab.

r3dm1ke:
### Description
This PR removes the SEO permission from the visibility tab and adds null checks to a few components.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/.chachalog/_u2yAwRr.md
+++ b/.chachalog/_u2yAwRr.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": patch
+---
+
+Removed seo permission from visibility tab and hardeded components for nulls (#2289)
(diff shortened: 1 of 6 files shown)


gflores-jahia (commented):
Can we get a bit of context about how it seems to be working in CE previously, but then stopped working along the way ? 

Do we maybe need to replace it with something else instead of just removing?  I see some references to [viewVisibilityTab](https://github.com/search?q=org%3AJahia+viewVisibilityTab&type=code) in our code base and maybe we can replace it with this permission instead ?

gflores-jahia (approved):

PR #2292, opened 2026-04-28
fix: start typing contextJsParameters

GauBen:
The `window.contextJsParameters` var is starting to be typed following https://github.com/Jahia/javascript-components/releases/tag/%40jahia%2Fui-extender%401.3.0

Many things are missing, this is just the beginning, let's do it step by step

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "@jahia/icons": "^1.1.2",
     "@jahia/moonstone": "^2.17.3",
     "@jahia/react-material": "^3.0.5",
-    "@jahia/ui-extender": "^1.1.1",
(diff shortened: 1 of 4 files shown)


gflores-jahia (approved):

gflores-jahia (commented):
I also see a few more references, are they maybe ok to update too, or just leave as-is for now? https://github.com/search?q=repo%3AJahia%2Fjcontent+contextJsParameter&type=code&p=1

GauBen:
We have plenty, I only did the ones that would get typing (`contextJsParameters.config` is currently unknown, kept it as is)

PR #2295, opened 2026-04-29
fix: Allow to have target attribute in the tooltip link

### LINKED ISSUE #2054: Provide more formating possibilities in UI tooltips

AKarmanov:
### Description
Allow to have target attribute in the tooltip link so customers can open the link in a new tab.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/Sanitizer.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/Sanitizer.java
@@ -13,7 +13,7 @@ public class Sanitizer {
         .allowCommonInlineFormattingElements()
         .allowUrlProtocols("http", "https")
         .allowElements("a")
-        .allowAttributes("href").onElements("a")
+        .allowAttributes("href", "target").onElements("a")
         .toFactory();
 
     public static final PolicyFactory COMMON_FORMATTING = Sanitizers.FORMATTING;


gflores-jahia (approved):

gflores-jahia (commented):
From previous discussion, can we also white list with `.allowAttributes("target").matching(true, "_blank").onElements("a")` ?

AKarmanov:
> From previous discussion, can we also white list with `.allowAttributes("target").matching(true, "_blank").onElements("a")` ?

We could discuss with the team but I don't know if it will give us any security benefits.

Issue #2029, opened 2025-10-08
Content that is not created on the page shouldn't be editable in Page Builder


ffffffelix:
### Context / Problem to solve / Design
Currently, when I use simple content or content with children as a content reference, we don't have the same behavior.
When I use a content item with children as a reference, I can edit/move all sub-nodes in Page Builder (and Page Composer). This modification impacts all instances where this content is used, so it can have a huge impact on the site. it's very confusing for users.

In the structured view, we only display "reference content" and not the items included in the reference.

### Solution
Content items (and sub content) used as reference must not be hoverable/selectionable/editable in Page Builder. The only way to edit it must be to go to the source content.
The same principle applies to content rendered from a query; such items should only be editable from the source.

### Story
As an editor and chief
I want to prevent my contributors from editing content that can impact other places without them being aware of what they do.

## Testable scenarios

**Setup**
Given I create a content reference in a page with Page Builder
And I pick a content list with items

**Case 01: hovering an item used in a content reference**
When I am hovering over an item in the referenced list
Then a dashed border is displayed around the content reference 

**Case 02: hovering over an item rendered from a content type query**
When I am hovering over an item rendered from a query
Then a dashed border is displayed around the query item

**Case 03: hovering an item rendered from a template query**
When I am hovering over an item rendered from a query built into the template
Then nothing happened

**Case 04: double click on an item used in a content reference**
When I double-click on an item in the referenced list
Then Content Editor is displayed to edit the content reference node

**Case 05: double click on an item rendered from a content type query**
When I double-click on an item rendered from a query
Then Content Editor is displayed to edit the query item

**Case 06: double click on an item rendered from a template query**
When I double-click on an item rendered from a query built into the template
Then nothing happened

**Case 07: Add action "Go to source"**
When I right-click on a content reference
Then the action "Go to source" is available in the contextual menu
And the action "Edit reference source" is removed

### Test Strategy
Cypress tests?

### Labels 
En: Go to source
Fr: Aller à la source
De:

AKarmanov:
I added translation, I have a fix for query issue but it has a limitation, it will not work for nodes from the same page as I simply look for parent path to not be available in nodes. I did not add it to the PR.

However, I also noticed that this issue only happens when you look for jnt:content, specific content jnt:news etc. doesn't have this issue as it renders content directly and does not create a module. I also feel like this could be an edge case as I'm not sure how many people will have such a broad query targeting jnt:content.

[...]

PR #2300, opened 2026-05-04
fix: Render box and placeholder for absolute areas

AKarmanov:
### Description
Render box and placeholder for absolute areas. We have [previously](https://github.com/Jahia/jcontent/issues/2029) implemented a check to prevent modules that originate from other sites to be editable in Page Builder. In some cases the check also excluded absolute areas originating at the root of the site, which made them uneditable. This pr fixes how absoluteAreas behave.
#### Source code

#### Tests

> [!TIP]


tests/provisioning-manifest-snapshot.yml
```diff
@@ -14,6 +14,8 @@
     - 'mvn:org.jahia.modules/jahia-ui-root'
     - 'mvn:org.jahia.test/jcontent-test-module'
     - 'mvn:org.jahia.test/jcontent-test-template'
+    - 'mvn:org.jahia.modules/javascript-modules-engine'
```
gflores-jahia: I don't think this is needed as it's already part of core modules

gflores-jahia (commented):
Works as designed, but there is a weird behavior in that we're not able to edit the links added to the absolute area `aaSiteRootLinks` with site as parent. 

This makes sense based on our rules as the area is not a descendant of any page, but I think needs some discussion in terms of usability. Maybe we need to allow all absolute areas?

AKarmanov:
> Works as designed, but there is a weird behavior in that we're not able to edit the links added to the absolute area `aaSiteRootLinks` with site as parent.
> 
> This makes sense based on our rules as the area is not a descendant of any page, but I think needs some discussion in terms of usability. Maybe we need to allow all absolute areas?

I didn't notice/try it. Not sure if it makes sense, I guess not if you look at it from page composer perspective. We can edit those links in page composer. It would be better if we dropped this isDescendant check altogether, there is very little gain for all the potential issues we might have because of it. We can weaken it to check if the node is originating from same site.

gflores-jahia (dismissed):
Approving for now, not sure if it makes sense to discuss this with this PR or needs to be taken as a separate issue. But I'll let you decide Alex if you'd want to merge it now or wait to clarify.

AKarmanov:
> Approving for now, not sure if it makes sense to discuss this with this PR or needs to be taken as a separate issue. But I'll let you decide Alex if you'd want to merge it now or wait to clarify.



> Approving for now, not sure if it makes sense to discuss this with this PR or needs to be taken as a separate issue. But I'll let you decide Alex if you'd want to merge it now or wait to clarify.

I simplified the check to only consider nodes from the same site as we talked in the daily.

gflores-jahia (approved):

PR #2302, opened 2026-05-05
chore: Use local resources instead of the one provided by assets module

dgriffon:
Fixes https://github.com/Jahia/jahia-private/issues/4800
### Description

This removes the usage of the assets resource  `/modules/assets/icons/jnt_contentList.png`.

### Notes

That icon is used in the toolbar header of page builder, it is a fallback in case of a nodetype icon not resolved for a given area if the content is of type `list`  (resolved in [ModuleTag](https://github.com/Jahia/jahia-private/blob/0d95959dfe7d74a6facbcc678e5e3e3f45666b6c/taglib/src/main/java/org/jahia/taglibs/template/include/ModuleTag.java#L594)).

The usage has been removed as not used anymore (see [comment](https://github.com/Jahia/jcontent/pull/2302#issuecomment-4380478805))

### Testing
No missing icon after the change. List icon should be displayed for contentList.


gflores-jahia:
For context, change around resolution of header icon was done as part of this PR https://github.com/Jahia/jcontent/pull/1889/changes

For the logic condition, even though it's meant to be a default catch-all for list type, it's never reached as nodetypes ultimately resolve to nt:base which provides a default icon. We can either keep it in case logic for providing default icons changes in the future (unlikely), or completely remove it as it's currently dead code. 

I would actually be in favour of just removing it completely but @AKarmanov @dgriffon not sure if you have thoughts on this.

dgriffon:
> For context, change around resolution of header icon was done as part of this PR https://github.com/Jahia/jcontent/pull/1889/changes
> 
> For the logic condition, even though it's meant to be a default catch-all for list type, it's never reached as nodetypes ultimately resolve to nt:base which provides a default icon. We can either keep it in case logic for providing default icons changes in the future (unlikely), or completely remove it as it's currently dead code.
> 
> I would actually be in favour of just removing it completely but @AKarmanov @dgriffon not sure if you have thoughts on this.

I agree on removing the code, I was about to do so, thank you for the confirmation!

GauBen:
Yay! 🎉

gflores-jahia (approved):

Issue #2168, opened 2026-02-11
Investigate cypress test crashes for some test suites


gflores-jahia:
More instances of test crashes when running integration tests happening on all workflows (on-change, merge to master, and nightlies). Output is the following for example:

```
  cypress             |   Running:  createContentI18N.cy.ts                                                        (6 of 20)
  cypress             | 
  cypress             | 
  cypress             |   Create content tests in I18N site
  cypress             |     ✓ Can create work in progress content for en/fr properties (25855ms)
  cypress             |     ✓ keeps "create another" checkbox state when switching languages  (18137ms)
  cypress             |     ✓ Can create a news in en -> fr and then create a new one in en only  (38429ms)
  cypress             | 
  cypress             | We detected that the Chrome Renderer process just crashed.
  cypress             | 
  cypress             | We have failed the current spec but will continue running the next spec.
  cypress             | 
  cypress             | This can happen for a number of different reasons.
  cypress             | 
  cypress             | If you're running lots of tests on a memory intense application.
  cypress             |   - Try increasing the CPU/memory on the machine you're running on.
  cypress             |   - Try enabling experimentalMemoryManagement in your config file.
  cypress             |   - Try lowering numTestsKeptInMemory in your config file during 'cypress open'.
  cypress             | 
  cypress             | You can learn more here:
  cypress             | 
  cypress             | https://on.cypress.io/renderer-process-crashed
```

Following tests affected at the moment:

- selectorTypes/textFieldInitializerTest
- selectorTypes/textRegexInDefinition
- contentEditor/createContent
- contentEditor/createContentI18N
- contentEditor/pageAsModel
- contentEditor/workInProgress
- contentEditor/usagesScreen

Ran all locally and tests are passing.

Notes:

- selectorTypes tests still running on page composer; can be rewritten with jcontent instead
- 3 suites are only running with one test. See if long running has impact
- see if possible to do some memory profiling
- see options if need to increase resources on test env
- not sure if specific tests are causing issues, or that the timing of the runs are causing it to fail after some point

PR #2308, opened 2026-05-06
fix: recursively resolve transitive mixin extensions in getExtendMixins

dgriffon:
Fixes https://github.com/Jahia/jahia-private/issues/4674
### Description
Fixed **getExtendMixins** to return the full transitive closure of mixin extensions instead of only direct extensions.
Replaced single-pass loop with a queue-based iterative approach, preventing duplicates and infinite loops via a visited set.

#### Source code

#### Tests

Cypress test were not provided as the current test regarding addMixin are [skipped](https://github.com/Jahia/jcontent/blob/78eb8705a97628c8d9f51bd523d4b0756232e6a3/tests/cypress/e2e/selectorTypes/addMixinChoicelistInitializers.cy.ts#L7)
Test was skipped in context this [issue](https://github.com/Jahia/jcontent/pull/2162) see the [comment](https://github.com/Jahia/jcontent/pull/2162#issuecomment-3899220349)
And should be addressed by https://github.com/Jahia/jcontent/issues/2168
(note that the test itself is not part of the list)
> [!TIP]


gflores-jahia (commented):
Lgtm, thanks ! 

Could we maybe add the issue scenario as a cypress regression test case?

dgriffon:
> Lgtm, thanks !
> 
> Could we maybe add the issue scenario as a cypress regression test case?

Thank you, I added a cypress test and enable back the one that was skipped, please have a look to my implementation, it may be improved by techniques you know to use properly jContent page objects

tests/cypress/e2e/selectorTypes/addMixinChoicelistInitializers.cy.ts
```diff
@@ -29,112 +27,170 @@ describe.skip('Add Mixin by using choice list initializers (Image Reference)', (
...
+
+    it('Can create and edit a vanilla node and switch between transitive mixin link types', () => {
+        // Open content editor for mynt:nodetypeVanilla which has mymix:mixinExtendWithLink
+        // extending it, which itself extends mymix:commonLink — transitive resolution is required
+        const contentEditor = jcontent.createContent('cent:nodetypeVanilla');
+
+        contentEditor.toggleOption('cemix:mixinExtendWithLink');
+        const linkTypeField = contentEditor.getChoiceListField('cemix:mixinExtendWithLink_linkType');
+
+        contentEditor.getTitle().should('be.visible').and('contain', 'Create');
```
gflores-jahia: should this one be at the beginning before we start doing anything in CE ?

gflores-jahia (changes requested):
@dgriffon I just checked this one and the test you've added is passing at least locally for me. What I'd recommend is add the `.skip` back to the original test, as it's currently crashing the run, and try to run only the test you just added for now and see if that's working.

tests/cypress.config-selectorTypes.ts
```diff
@@ -1,6 +1,6 @@
 import {defineConfig} from 'cypress';
 import {baseConfig} from './cypress.config.common';
 
-baseConfig.e2e.specPattern = ['cypress/e2e/selectorTypes/**/**.cy.ts'];
+baseConfig.e2e.specPattern = ['cypress/e2e/selectorTypes/**/addMixinChoicelistInitializers.cy.ts'];
```
gflores-jahia: We need to revert this back to restore tests

.github/workflows/on-code-change.yml
```diff
@@ -85,7 +85,6 @@ jobs:
       fail-fast: false
       matrix:
         config_file:
-          - cypress.config-contentEditor.ts
```
gflores-jahia, line 89: We need to revert to restore tests.

gflores-jahia (commented):
Thanks for fixing the choicelist test. looks ok on my side (although still waiting qa feedback). In the meantime, we'll also probably need to restore the excluded tests before merging.

gflores-jahia (dismissed):

baptistegrimaud:
> Thanks for fixing the choicelist test. looks ok on my side (although still waiting qa feedback). In the meantime, we'll also probably need to restore the excluded tests before merging.

@gflores-jahia do you happen to know where we're at on this task?
Are we still waiting for some QA feedback/confirmation?
Thanks!

gflores-jahia:
@baptistegrimaud I think code-wise we are ok, but seeing that we have some failures and had a few major merges since, I'd recommend maybe syncing this PR branch up to latest main and re-run just to check how tests look.

@sdusolle I don't know if you're ok verifying the added tests here 🙏

sdusolle:
The tests look OK on my side, but there are some failures included the new mixin test: 
`- FAIL: Can create and edit a vanilla node and switch between transitive mixin link types`

x0h01 (approved):

TheKlem:
**Claude analysis** (run by @TheKlem, verified on a running Jahia 8.2.3.2 — not a code read)

No objection to the change itself; one thing worth separating before merge.

## This PR does not fix #2746, which currently links to it

I deployed the scenario from #2746 on a real instance — three player mixins, each `extends = videoPlayer`, each inheriting `image` from a shared `genericPlayer` supertype — and asked the server what form it generates, so the UI and the choicelist initializer are out of the picture:

```
youtubeGenericPlayer      fields: youtubeId, image
vimeoGenericPlayer        fields: vimeoId, vimeoVideoAccessKey     <- image missing
switchTubeGenericPlayer   fields: switchTubeId                     <- image missing
```

Then built jcontent with this PR's `EditorFormServiceImpl` on top of current main, deployed, and re-ran the same query: **byte-identical output**. No difference at all.

That is consistent with what the code does rather than a surprise. Discovery is not the failing step in that scenario — all three player mixins are already found without this PR, which is why all three fieldsets render. What goes missing is one inherited *field* inside two of them, and that happens later, in `FieldSet.mergeFields` → `Form.findAndRemoveField`, which *removes* a field from wherever it already lives so it can be re-homed. A field name can exist in only one fieldset per form, so siblings sharing an inherited property fight over one copy. Details and the exact lines are in [my comment on #2746](https://github.com/Jahia/jcontent/issues/2746#issuecomment-5571146294).

## On the PR itself

The BFS looks right, the `visited` set genuinely prevents both duplicates and cycles in diamond hierarchies, and pulling `isExtensionAvailable` out is an improvement on the original condition. It is plausibly correct for the case the description names (`internalLink extends commonLink extends nodetypeVanilla`) and for jahia-private#4674.

Two suggestions:

1. **Unlink it from #2746**, so that ticket does not get closed by this merge — the reported symptom survives it.
2. The reviewer asked twice for a Cypress case covering the issue scenario. That is worth insisting on here specifically: a test written against #2746's definitions would have shown this PR leaving the symptom in place, which is the whole reason this comment exists. `getExtendMixins` returning more mixins is hard to observe from the outside; the fieldsets and their fields are not.

PR #2313, opened 2026-05-12
fix: Correctly resolve insertion point buttons.

### LINKED ISSUE #2310: PB wrong behavior for child nodes specified in definition when only one child

reproduced with current 3.7 by Cedric
works in Page Composer, not in PB
https://support.jahia.com/browse/STSCI-624

### Steps to reproduce

in the client tocket

### Current Behavior

child item are not creatable

### Desired Behavior

user should be able to cretae the child item

AKarmanov:
### Description
Correctly resolve insertion point buttons. Make sure DOMs placeholder elements are used as the source of truth for driving what types of create buttons are available. Make sure that create buttons and insertion points are shown correctly for definitions with fixed and variable number of children. So if I have a definition with 4 different types of children defined as `+ *` I can see 4 distinct buttons for these types and not 4 buttons of the same type.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/ContentEditor/actions/jcontent/createContent/createContentActionPB.jsx
+++ b/src/javascript/ContentEditor/actions/jcontent/createContent/createContentActionPB.jsx
@@ -46,7 +46,8 @@ export const CreateContentPB = ({
         let resolvedNodeTypesForAction = [];
         // NodeTypes come from placeholder's nodetypes attribute, as each action is created for such placeholder we want to match what that button can create
         if (resNode.acceptedNodeTypes.size > 0) {
-            if (nodeTypes?.length > 0) {
+            // If nodetypes is defined, it will be used as the source of truth for calculating available note types
+            if (nodeTypes) {
                 resolvedNodeTypesForAction = nodeTypes
                     .map(nt => resNode.acceptedNodeTypes.get(nt))
                     .filter(nt => nt !== undefined);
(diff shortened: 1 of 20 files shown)


src/javascript/ContentEditor/actions/jcontent/createContent/createContentActionPB.jsx
```diff
@@ -104,7 +112,8 @@ CreateContentPB.propTypes = {
     loading: PropTypes.func,
     onVisibilityChanged: PropTypes.func,
     labelProps: PropTypes.object,
-    nodeData: PropTypes.object
+    nodeData: PropTypes.object,
+    nodeTypes: PropTypes.object
```
gflores-jahia: Incorrect proptype here. should be array of strings (parsed from create.jsx:55).

gflores-jahia (commented):
lgtm. just minor proptype issue to be resolved.

src/javascript/ContentEditor/actions/jcontent/createContent/createContentActionPB.jsx
AKarmanov: ahh, yes, good catch!

gflores-jahia (approved):
looks good. thank you for the comment explanation in create.

PR #2355, opened 2026-06-03
test: add Cypress tests

sdusolle:
### Description
Move Selenium tests to Cypress:
- testCreateContentAction
- testAdditionalAccordion
- extend coverage on contentFolder tests
- also removed Digitall website from the existing tests

#### Source code

#### Tests

> [!TIP]


tests/cypress/e2e/jcontent/additional.cy.ts
```diff
@@ -0,0 +1,49 @@
...
+        JContent.visit(siteKey, 'en', 'apps');
+
+        cy.get('header[aria-controls="apps"]')
+            .closest('section')
+            .should('contain', 'Link checker');
+        cy.get('header[aria-controls="apps"]')
+            .closest('section')
+            .should('contain', 'SEO');
+
+        clickAppsTreeItem('linkchecker');
```
gflores-jahia: We also already have Accordion and AccordionItem page objects in jcontent e.g. `getAccordionItem()` so we can maybe reuse or enhance that instead.

tests/cypress/e2e/jcontent/additional.cy.ts
```diff
@@ -0,0 +1,49 @@
...
+            .find('h2')
+            .contains('Link Checker')
+            .should('be.visible');
+    });
+
+    it('should not display additional panel on System Site', () => {
+        const jcontent = JContent.visit(siteKey, 'en', 'apps');
+        jcontent.getSiteSwitcher().select('System Site');
+        cy.get('header[aria-controls="apps"]', {timeout: 5000})
+            .should('not.exist');
```
gflores-jahia: We also already have Accordion and AccordionItem page objects in jcontent e.g. `getAccordionItem()` so we can maybe reuse or enhance that instead.

tests/cypress/page-object/contentFolder.ts
```diff
@@ -0,0 +1,39 @@
...
+    constructor(jcontent: JContent, site: string, language: string, parentPath: string, name: string) {
+        super();
+        this.jcontent = jcontent;
+        this.site = site;
+        this.language = language;
+        this.parentPath = parentPath;
+        this.name = name;
+    }
+
+    static create(jcontent: JContent, site: string, language: string, parentPath: string, folderName: string): ContentFolder {
```
gflores-jahia: I appreciate the effort to code for reusability here and it's very welcome, but I'm seeing that it could be improved if refactored as a different level of abstraction.

imo, other than the `create` part which is specific to creating folders, the other functions have a lot more similarity to a `ContentTableRow` object. So maybe, can be split two steps:

- Refactor `create` to `createFolder(folderName: string): ContentTableRow` in `jcontent.ts`. This also has an additional benefit of implicit check that the folder was created and exists as a ContentTableRow to be returned, and can also replace the check steps in some of the content folder tests.
- Move `markForDeletion(): ContentTableRow` and `deletePermanently(): ContentTableRow` helper methods to ContentTableRow. This makes these action helpers be more widely available in Rows in general than just specific to content folders.

Something to consider is maybe create `CreateFolderDialog` instead that can provide functions for interacting with that dialog or verifying states e.g. duplicate folder check.

Let me know if you think this works for your use case, or if you have other ideas or suggestions.

gflores-jahia (commented):
Thank you Sonia for the test PR as always. I've added a few suggestions. Lmk if that makes sense.

gflores-jahia (dismissed):
Looks good, thank you for the changes !

brigitteradl (dismissed):

brigitteradl (approved):

gflores-jahia (approved):

PR #2363, opened 2026-06-05
feat: Add loading state for context menus

### LINKED ISSUE #2251: Contextual menu takes up to 2s before being displayed
- Jahia 8.2.2.0
- jContent 3.6.0

### Description

When right-clicking an element, there is a very noticeable 1-2s delay before the contextual menu is shown. It doesn’t seem that bad at first, but after spending some time editing elements on pages, it’s infuriating to have to wait every time before being able to click on edit.

It gives a very laggy/heavy vibe to the whole UI, making working extensively in edit mode an unpleasant experience for the editors. “It’s slow” is what we heard the most.

### Notes
This happens on empty pages too.

If the reason of the delay is that certain options of the right-click menu require some processing, it’d be way better to display the menu without delay on right-click, and then render the options that need more processing.

gflores-jahia:
### Description

Partially fixes issue around noticeable delay in context menus before it appears. This is due the bulk graphql query for determining visibility logic and permission checks for all actions, and that we only display context menu until response is received.

#### Performance fixes

 - Show menu immediately — the context menu now opens at once with skeleton loading placeholders instead of waiting for all actions to resolve
   - MenuLoadingContext — new React context propagates loading state; MenuItemRenderer renders <MenuItemSkeleton> while actions are resolving, then atomically replaces them when ready
 - loadingTypes fix in createContent.utils.js — changed to loading && !data so cached data is used immediately on subsequent opens; added notifyOnNetworkStatusChange: true to correctly track background refetch
 - Pre-gate visibility checks (nodeVisibilityUtils.js) — 20+ actions now call isDefinitelyHidden() to return early before triggering GQL, using hideOnNodeTypes, hideMixins, and showOnNodeTypes checks on data already available in the node payload
   - The function is intentionally pessimistic: it only returns true (hidden) when it can be determined with certainty from the available node data — if in any doubt, it returns false and falls through to the full GQL resolution. This is esp true for node types, where a full hierarchy is needed to check so we only enforce prechecks only on exact primary type matching, otherwise we defer to the original useNodeChecks.
 - Moonstone update — bumped to 2.20.0 which includes a ResizeObserver-based usePositioning fix, preventing menu repositioning flicker after loading transition
 
#### Test coverage

 - contextMenuVisibility.cy.ts — 7 Cypress tests covering skeleton loading state and per-node-type action visibility (jnt:file, jnt:contentFolder, jmix:markedForDeletion)**

### DIFF
--- a/src/javascript/JContent/actions/compareHtmlAction/compareHtmlAction.jsx
+++ b/src/javascript/JContent/actions/compareHtmlAction/compareHtmlAction.jsx
@@ -3,12 +3,15 @@ import PropTypes from 'prop-types';
 import {createEncodedHashString} from '../../CompareDialog/util';
 import {useNodeChecks} from '@jahia/data-helper';
 import {useSelector} from 'react-redux';
+import {isDefinitelyHidden} from '../utils/nodeVisibilityUtils';
 
-export const CompareHtmlActionComponent = ({path, render: Render, loading: Loading, ...others}) => {
+export const CompareHtmlActionComponent = ({path, node: prefetchedNode, render: Render, loading: Loading, ...others}) => {
     const language = useSelector(state => state.language);
+    const skip = isDefinitelyHidden(prefetchedNode, {showOnNodeTypes: ['jnt:page', 'jmix:mainResource']});
     const res = useNodeChecks(
         {path, language},
         {
+            skip,
             showOnNodeTypes: ['jnt:page', 'jmix:mainResource'],
             getAggregatedPublicationInfo: {subNodes: true}
         }
@@ -21,6 +24,10 @@ export const CompareHtmlActionComponent = ({path, render: Render, loading: Loadi
 
     if (res.loading) {
         return (Loading && <Loading {...others}/>) || false;
+    }
+
+    if (skip) {
+        return false;
     }
 
     const isVisible = res.checksResult && (res?.node?.aggregatedPublicationInfo.publicationStatus === 'PUBLISHED' ||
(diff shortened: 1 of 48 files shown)


sdusolle (dismissed):

src/javascript/JContent/actions/compareHtmlAction/compareHtmlAction.jsx
```diff
@@ -3,12 +3,15 @@ import PropTypes from 'prop-types';
 import {createEncodedHashString} from '../../CompareDialog/util';
 import {useNodeChecks} from '@jahia/data-helper';
 import {useSelector} from 'react-redux';
+import {isDefinitelyHidden} from '../utils/nodeVisibilityUtils';
 
-export const CompareHtmlActionComponent = ({path, render: Render, loading: Loading, ...others}) => {
+export const CompareHtmlActionComponent = ({path, node: prefetchedNode, render: Render, loading: Loading, ...others}) => {
     const language = useSelector(state => state.language);
+    const skip = isDefinitelyHidden(prefetchedNode, {showOnNodeTypes: ['jnt:page', 'jmix:mainResource']});
```
AKarmanov: Maybe we can make these params a const (showOnNodeTypes etc.) since they are shared with useNodeChecks?

gflores-jahia: Makes sense. I went through all the actions and refactored the shared objects as variables. I had to rebase after a conflict but should be the same. The last commit should have these fixes.

r3dm1ke (approved):

PR #2366, opened 2026-06-05
fix: disable create another checkbox if limit reached

### LINKED ISSUE #1962: "Create another" checkbox bypasses j:numberOfItems constraint in lists

Jahia version: 8.1.x
Affected configuration: j:numberOfItems on jnt:area

### Steps to reproduce

Define a list area in a content type with a maximum number of allowed items, e.g.:
```
<myarea j:numberOfItems="2" jcr:primaryType="jnt:area"/>
```
Open the content in edit mode and start adding items to the list.

When creating the first item, check the "Create another" checkbox.

Continue creating items using this mode.

Observe that it is possible to exceed the defined maximum number of elements (in this case, 2).

Once the checkbox is unchecked and the limit is reached, the system correctly prevents adding new items.

### Current Behavior

The j:numberOfItems constraint is not enforced when using the "Create another" checkbox.
Users are able to create more items than the configured maximum.
Once "Create another" is unchecked and the limit has been reached, the UI behaves correctly and blocks further additions.

### Desired Behavior

The "Create another" checkbox should respect the j:numberOfItems constraint and not allow creation of additional items beyond the limit.
Ideally, the checkbox should be hidden or disabled automatically when the maximum is reached, even while still in "create another" mode.

r3dm1ke:
### Description
This PR adds a check to the create another checkbox in CE if child node limit is reached.

#### Source code

#### Tests

> [!TIP]


gflores-jahia (commented):
Thanks for the PR. It looks well thought out and I appreciate the optimizations, especially the useMemo on the editor context.

  However, I have an overall design concern about whether `disableCreateAnother` belongs at the CE context level, considering it's a create-only mechanism. Two suggestions as an alternate design:

   - Context placement: `disableCreateAnother` could be computed at the lower Create.jsx level and threaded through the existing createAnother prop by extending it with a disabled field, rather than adding it to the shared CE context.
   - Query consolidation: The required fields could be included directly in CreateFormQuery (the parent node is already queried there). Since the query is already refetched on each "create another" save, the data is always fresh — making the separate `useNodeInfo` call and the useRef/count session tracking both unnecessary.

It's a bit high-level, but let me know what you think or if I maybe missed some detail.

sdusolle (dismissed):

gflores-jahia (dismissed):
That was a quick turnaround, thanks ! Just need to check on the failing tests

sdusolle (approved):

gflores-jahia (approved):

PR #2379, opened 2026-06-08
test: Fix description tests, add skill

AKarmanov:
### Description
Fix description tests, add skill. The test is in graphql and it breaks if descriptions are missing. I added a skill so that descriptions are generate if AI is used.
#### Source code

#### Tests

> [!TIP]


gflores-jahia (approved):
Thank you for the fix !

PR #2389, opened 2026-06-09
fix: Update how nodetypes are calculated for insertion points

AKarmanov:
### Description
Update how nodetypes are calculated for insertion points when contribute types are defined on the node. 

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/EditFrame/InsertionPoints.jsx
+++ b/src/javascript/JContent/EditFrame/InsertionPoints.jsx
@@ -16,10 +16,10 @@ const getNodeTypes = e => {
 
     if (e.dataset.jahiaParent) {
         const parent = e.ownerDocument.getElementById(e.dataset.jahiaParent);
-        const parentNt = parent?.getAttribute('nodetypes');
-        if (parentNt) {
-            parentNt.split(' ').forEach(t => types.add(t));
-        }
+        // const parentNt = parent?.getAttribute('nodetypes');
+        // if (parentNt) {
+        //     parentNt.split(' ').forEach(t => types.add(t));
+        // }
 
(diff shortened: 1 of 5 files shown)


src/javascript/JContent/EditFrame/InsertionPoints.jsx
```diff
@@ -6,22 +6,44 @@ import {useSelector} from 'react-redux';
 import {usePasteData} from '~/JContent/EditFrame/Boxes/dataHooks/usePasteData';
 
 const getNodeTypes = e => {
```
gflores-jahia, line 33: I know there's already inline comments but as the logic for calculating these nodetypes are starting to get complicated. I think we should add a high-level summary explaining what's happening here, how it works and for which scenarios. (hopefully AI can help generate that).

gflores-jahia (commented):
Code looks ok to me but I think we need to maybe start adding some high level overview of the logic for calculating these nodetypes to make it (somewhat) easier to understand.

src/javascript/JContent/EditFrame/InsertionPoints.jsx
AKarmanov: I added a comment, hopefully it makes things a little clearer.

r3dm1ke (approved):

src/javascript/JContent/EditFrame/InsertionPoints.jsx
gflores-jahia: Very helpful and informative. I think the why is even more valuable here to add some context around the code and always the hardest to figure out after some time. So thanks for adding that.

gflores-jahia (approved):

sdusolle (approved):

PR #2393, opened 2026-06-10
fix: validation for fields that are empty initially

### LINKED ISSUE #2347: validators don't work as expected

8.2.3.0

### Steps to reproduce

Implement a custom validator or use osgi validator-examples (note: the example use old hibernate validators which are no longer working, so it must be changed to javax.validation.constraints). 

### Current Behavior

The validator works, but no validation message appears in the UI, only on the right bottom of the page do you get a common error message: 

Image

So contributors/editors doesn't know whats wrong. 
As an example you can also use the attached example (I only modified the NotEmpty validator on the "jnt:testContent" definition in the example. 

[validator-samples.zip](https://github.com/user-attachments/files/28350789/validator-samples.zip)

### Desired Behavior

A message should be displayed directly in the UI or on the field (as it was the case before).
Also a custom message would be nice when it is specified as validator property like:
`
@NotEmpty(message = "My error message")`

r3dm1ke:
### Description
This PR fixes the following problem: if a field has no default value, it will not show custom error message when validated.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/.chachalog/W5ookQ45.md
+++ b/.chachalog/W5ookQ45.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": patch
+---
+
+Fixed custom validation messages for fields with no default value.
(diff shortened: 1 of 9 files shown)


gflores-jahia (approved):
All good. I had to validate some of your changes in fields.utils didn't reverse https://github.com/Jahia/jcontent/pull/2322. In this case, `propsToSave/delete` was unchanged and only affected `propFieldNameMapping`. Also added test on `contentEditor/permissions.cy` is still passing so we should be good. Thanks !

sdusolle (approved):

PR #2410, opened 2026-06-11
fix: exclude supertype extend mixins from fieldset processing to prev…

### LINKED ISSUE #2412: addMixin fieldset properties non-deterministically assigned when two mixins share the same extends target

- Jahia: 8.2.x
- jContent: 3.6.0

### Steps to reproduce

1. Define two mixins where one extends the other, and both have `extends = jnt:page`:
```
[mymix:pageA] > jmix:templateMixin, mymix:pageHeader mixin
extends = jnt:page
- typeMarker (string) = 'pageA' autocreated protected

[mymix:pageHeader] mixin
extends = jnt:page
- title (string) internationalized
- description (string) internationalized
```

2. Register a `ModuleChoiceListInitializer` that returns a `ChoiceListValue` with `addMixin = mymix:pageA`.
3. Create a page using the template mapped to `mymix:pageA`.
4. Observe whether `title` and `description` appear in the Content Editor form.
5. Repeat several times or on different environments.

### Current Behavior

`getExtendMixins()` in `EditorFormServiceImpl` returns **both** `mymix:pageHeader` and `mymix:pageA` because both have `extends = jnt:page`. The method iterates over a `HashMap`, so processing order is non-deterministic.

- If `mymix:pageA` is processed first: its supertypes (including `mymix:pageHeader`) are added to `processedNodeTypes`, so `mymix:pageHeader` is skipped. Only one fieldset exists (`mymix:pageA`) with all properties. Fields show correctly.
- If `mymix:pageHeader` is processed first: both fieldsets are created. During merge, `findAndRemoveField()` moves `title` and `description` from the `mymix:pageA` fieldset into the `mymix:pageHeader` fieldset. Since `addMixin` only activates `mymix:pageA`, the `mymix:pageHeader` fieldset is never activated and its fields never appear.

Result: fields from the supertype mixin show intermittently depending on JVM HashMap iteration order.

## Desired Behavior

When a mixin `B` is both a supertype of mixin `A` and independently present in the extend mixins list, `getExtendMixins()` (or the subsequent processing loop) should detect the supertype relationship and skip `B`, ensuring only `A` is treated as the active extend mixin. Field assignment should be deterministic regardless of iteration order.

pvollenweider:
### Description

When two mixins both declare `extends = jnt:page` (or any shared target type) and one is a supertype of the other, `getExtendMixins()` returns both. Because the underlying collection iterates over a `HashMap`, processing order is non-deterministic. Depending on which mixin is processed first, `findAndRemoveField()` during fieldset merge moves properties from the correct activated fieldset into an inactive one, causing fields to silently disappear from the Content Editor form.

**Fix:** after collecting the extend mixin list, filter out any entry that is a supertype of another entry in the same list. The more specific mixin already carries all supertype properties via inheritance, so the supertype entry is redundant.

Discovered while investigating a customer case (ARKEMA-1813) where `akmmix:sampleCartPage > jmix:templateMixin, akmmix:sampleCartHeader` — both having `extends = jnt:page` — caused `title` and `description` from `akmmix:sampleCartHeader` to intermittently not appear in Content Editor after the `addMixin` mechanism applied `akmmix:sampleCartPage`.

#### Source code

#### Tests

**Manual validation:** define two mixins where one extends the other and both have `extends = jnt:page`. Map the more specific mixin via `addMixin`. Create a page using that template and confirm all inherited properties appear in Content Editor consistently across multiple attempts.

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
@@ -359,6 +359,12 @@ public class EditorFormServiceImpl implements EditorFormService {
             }
         }
 
+        // If mixin A extends mixin B and both are in the list, remove B.
+        // A already carries B's properties via the supertype chain, so keeping both
+        // causes non-deterministic field assignment during fieldset merge.
+        res.removeIf(mixin -> res.stream()
+            .anyMatch(other -> !other.getName().equals(mixin.getName()) && other.isNodeType(mixin.getName())));
+
         return res;
     }
 


gflores-jahia:
Recreated in https://github.com/Jahia/jcontent/pull/2467 as workflows don't run on forked repos.

Closing this PR.

PR #2416, opened 2026-06-12
fix: open-ended number boundary validation.

### LINKED ISSUE #2206: Number range validation fails for open-ended constraints like (0,) (invalidRange returned for valid LONG values)

jContent 3.2.0+ (issue suspected to be introduced around this version)
Custom CND definition with numeric range constraint

### Steps to reproduce

1. In a Jahia 8.2.x environment with jContent ≥ 3.2.0
2. Define a property with an open-ended numeric constraint:
```
- max (long) < '(0,)'
```
3. Deploy the module
4. Edit a node containing this property in Content Editor
5. Enter a valid integer value, e.g. 3
6. Click Save

### Current Behavior

The field is always marked as invalid with error `invalidRange`
Even valid integer values such as 1, 2, 3, etc., are rejected.
Cause (analysis):

In `rangeFieldValidation`, both lower and upper boundaries are evaluated unconditionally (file https://github.com/Jahia/jcontent/blob/master/src/javascript/ContentEditor/validation/validate.js )
```
const upperBoundaryOk = disableUpperBoundary
    ? fieldValueAsNumber < Number(upperBoundary)
    : fieldValueAsNumber <= Number(upperBoundary);
```
For an open-ended constraint `(0,)`:
- `upperBoundary` is `undefined`
- Number(undefined) returns `NaN`
- Any comparison with `NaN` returns `false`

Therefore, `upperBoundaryOk` is always `false`, and the validation always fails.

Unlike `dateFieldValidation`, the numeric validation does not skip boundary checks when a boundary is undefined.

Workaround:
Defining an explicit upper bound makes it work:
```
- max (long) < '(0,9223372036854775807]'
```
This may be related to this commit https://github.com/Jahia/jcontent/commit/50edea1b6382adea12915d798e8553a22d6e30ba

### Desired Behavior

Open-ended numeric constraints such as:
```
(0,)
[0,)
(,10]
```
should be handled correctly.

If upperBoundary (or `lowerBoundary`) is not defined, the validator should skip that boundary check instead of evaluating it against NaN.

The behavior should be consistent with `dateFieldValidation`, which only applies boundary checks when the boundary exists.

Valid values like `3` should be accepted for `(0,)` without requiring an artificial explicit upper bound.

### Proposed fix:

```
return fieldValues.every(fieldValue => {
    const fieldValueAsNumber = Number(fieldValue);

    // Parse bounds safely (open-ended ranges have undefined bounds)
    const parseBound = b => {
        if (b === undefined || b === null) { return undefined; }
        const s = String(b).trim();
        if (!s) { return undefined; }
        const n = Number(s);
        return Number.isNaN(n) ? undefined : n;
    };

    const lower = parseBound(lowerBoundary);
    const upper = parseBound(upperBoundary);

    // If a bound is missing, that side is considered OK (open-ended range)
    const lowerBoundaryOk = (lower === undefined) ? true :
        (disableLowerBoundary ? fieldValueAsNumber > lower : fieldValueAsNumber >= lower);

    const upperBoundaryOk = (upper === undefined) ? true :
        (disableUpperBoundary ? fieldValueAsNumber < upper : fieldValueAsNumber <= upper);

    return lowerBoundaryOk && upperBoundaryOk;
});
```

r3dm1ke:
### Description
This PR fixes the number validation with open-ended boundaries.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/.chachalog/JeUrmktv.md
+++ b/.chachalog/JeUrmktv.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": patch
+---
+
+Fixed validation of numbers when their range is open-ended.
--- a/src/javascript/ContentEditor/validation/validate.js
+++ b/src/javascript/ContentEditor/validation/validate.js
@@ -107,13 +107,26 @@ const rangeFieldValidation = (values, field) => {
             try {
                 const {lowerBoundary, disableLowerBoundary, upperBoundary, disableUpperBoundary} = extractRangeConstraints(constraint.value.string);
 
+                // Open-ended ranges like (0,) have an empty boundary; skip the check for that side
+                const parseBoundary = boundary => {
+                    if (boundary === undefined || boundary === null || String(boundary).trim() === '') {
+                        return undefined;
+                    }
+
+                    const boundaryAsNumber = Number(boundary);
+                    return Number.isNaN(boundaryAsNumber) ? undefined : boundaryAsNumber;
+                };
+
+                const lower = parseBoundary(lowerBoundary);
+                const upper = parseBoundary(upperBoundary);
+
                 return fieldValues
                     .every(fieldValue => {
                         const fieldValueAsNumber = Number(fieldValue);
                         // Lower boundary Check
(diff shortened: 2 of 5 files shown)


r3dm1ke:
The duplication issue in tests - do not need to fix IMO, they are very small and readable.

gflores-jahia (approved):

sdusolle (approved):

PR #2418, opened 2026-06-16
fix: Show error if no item is found in picker field.

### LINKED ISSUE #2350: ItemNotFoundException when saving a page with a broken reference

pmdemo / jahiacom 
jcontent 3.6

### Steps to reproduce

edit a page (luxe or any template set)
go to seo section
add an image
publish

go to jcontent media
delete the image (publish deletion)

go back to the page 
edit 
change the title
save 

### Current Behavior

ItemNotFoundException in graphql

### Desired Behavior

TBD. 
Validation error because broken reference or the save should work

AKarmanov:
### Description
Show error if no item is found in picker field. The user needs to be alerted as hitting the save button will result in an error, so it is better to address the issue beforehand. 

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/ContentEditor/SelectorTypes/Picker/Picker.jsx
+++ b/src/javascript/ContentEditor/SelectorTypes/Picker/Picker.jsx
@@ -178,7 +178,7 @@ export const Picker = ({
     const {t} = useTranslation('jcontent');
     const {lang} = useContentEditorConfigContext();
     const [isDialogOpen, setDialogOpen] = useState(false);
-    const {setFieldValue, setFieldTouched} = useFormikContext();
+    const {setFieldValue, setFieldTouched, setFieldError} = useFormikContext();
 
     const parsedOptions = getOptions(field, inputContext);
     const pickerConfig = mergeDeep({}, DefaultPickerConfig, inputContext.selectorType.pickerConfig, parsedOptions.pickerConfig);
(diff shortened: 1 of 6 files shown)


src/javascript/ContentEditor/validation/validate.spec.js
```diff
@@ -1,6 +1,19 @@
...
+
+const emailConstraint = () => ({valueConstraints: [{value: {string: EMAIL_PATTERN}}]});
+const valueListConstraints = () => ({
+    valueConstraints: [
+        {value: {string: 'value0'}},
+        {value: {string: 'value1'}},
+        {value: {string: 'value2'}},
+        {value: {string: 'value3'}}
+    ]
+});
```
gflores-jahia: [nitpick] why functions and not just literal objects, considering it's not using any param ?

src/javascript/ContentEditor/SelectorTypes/Picker/Picker.jsx
```diff
@@ -202,6 +202,16 @@ export const Picker = ({
         }
     }, [dispatch, pickerConfig.defaultViewMode]);
 
+    useEffect(() => {
+        if (notFound) {
+            // Must clear callstack to allow formik's validate function to run first.
+            setTimeout(() => {
+                setFieldError(field.name, 'refNotFound');
+                setFieldTouched(field.name, true, false);
+            }, 0);
```
gflores-jahia, line 212: need a clearTimeout cleanup here

AKarmanov: good catch

src/javascript/ContentEditor/validation/validate.spec.js
AKarmanov: I'll ask my AI, it's his creation

AKarmanov: Here's his answer: No good reason here. Since the object is never mutated by the tests, a plain const is fine and simpler. (:facepalm)

gflores-jahia (approved):

PR #2445, opened 2026-06-23
fix: Add constraints on BoxesContextMenu

AKarmanov:
### Description
Add constraints on BoxesContextMenu to prevent it from showing on jnt:page targets.
#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/EditFrame/Boxes/BoxesContextMenu.jsx
+++ b/src/javascript/JContent/EditFrame/Boxes/BoxesContextMenu.jsx
@@ -3,6 +3,58 @@ import {ContextualMenu} from '@jahia/ui-extender';
 import PropTypes from 'prop-types';
 import {useHoverContext} from '~/JContent/EditFrame/Boxes/HoverContext';
 
+// Right-clicks on a breadcrumb/dropdown item are handled specially (and never open the content menu).
+// Returns true when the event was a breadcrumb click and has been handled.
+const handleBreadcrumbFooter = event => {
+    if (!event.target?.closest('[data-sel-role="pagebuilder-itempath-dropdown"]')) {
+        return false;
+    }
+
+    event.preventDefault();
+    event.stopPropagation();
+    // Ctrl + click counts as a right click and triggers contextmenu; forward it as a real click.
+    if (event.ctrlKey) {
+        event.target.dispatchEvent(new MouseEvent('click', {
+            bubbles: true,
+            cancelable: true,
+            ctrlKey: true,
+            detail: 1,
+            screenX: event.screenX,
+            screenY: event.screenY,
+            clientX: event.clientX,
+            clientY: event.clientY
+        }));
+    }
+
+    return true;
+};
+
+// Resolve the element actually under the cursor at right-click time. Relying on the hovered path is
+// unsafe: it goes stale (e.g. when moving onto the page without a hover-out firing), which used to open
+// the menu for the wrong element or fall back to selecting the whole page. The box body overlay is
+// pointer-events:none, so event.target there is whatever shows through (sometimes a page gap with no
+// module). Inspecting the whole stack under the cursor pierces pointer-events:none, and the topmost
+// element resolving to a data-jahia-path wins (content modules and their box overlays both expose it).
+const resolveTargetPath = (frameDocument, event) => {
+    const stack = Array.from(frameDocument.elementsFromPoint?.(event.clientX, event.clientY) ?? [event.target]);
+    return stack
+        .map(element => element?.closest?.('[data-jahia-path]'))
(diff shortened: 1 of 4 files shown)


gflores-jahia:
Nice fix — the \`elementsFromPoint\` approach is solid and I can see why it was chosen. One thing I explored while reviewing: the root cause of the stale path appears to be in \`onMouseOut\` in \`Boxes.jsx\`. When the cursor leaves the iframe document (e.g. moves to the parent window), \`event.relatedTarget\` is \`null\`, which causes the entire clear-hover block to be skipped — including the \`clearTimeout\` — so the pending \`setTimeout\` from \`onMouseOver\` still fires and writes a stale path.

A minimal fix there would be:

\`\`\`js
// Boxes.jsx — onMouseOut
window.clearTimeout(timeout); // moved before the if

if (!event.relatedTarget ||  // null = cursor left the iframe, always clear
    (!isDescendantOrSelf(...) && !event.target.closest('#menuHolder'))
) {
    clearHovered();
    hoverProviderRef.current?.setHoveredPath(null);
}
\`\`\`

With hover reliable, \`BoxesContextMenu\` can read \`currentPathRef.current\` (same ref pattern already used for \`selection\`/\`nodes\`) and guard for \`null\` — no DOM traversal needed, and the \`selectionRef\`/\`nodesRef\` workaround can also be dropped once the \`useEffect\` returns a proper cleanup.

Happy to be wrong about the root cause if there are other stale scenarios I'm missing. Either way the \`elementsFromPoint\` approach works — just flagging this in case it simplifies things.

src/javascript/JContent/EditFrame/Boxes/BoxesContextMenu.jsx
```diff
@@ -3,6 +3,58 @@ import {ContextualMenu} from '@jahia/ui-extender';
...
+// pointer-events:none, so event.target there is whatever shows through (sometimes a page gap with no
+// module). Inspecting the whole stack under the cursor pierces pointer-events:none, and the topmost
+// element resolving to a data-jahia-path wins (content modules and their box overlays both expose it).
+const resolveTargetPath = (frameDocument, event) => {
+    const stack = Array.from(frameDocument.elementsFromPoint?.(event.clientX, event.clientY) ?? [event.target]);
+    return stack
+        .map(element => element?.closest?.('[data-jahia-path]'))
+        .find(Boolean)
+        ?.dataset.jahiaPath;
+};
```
gflores-jahia, line 44: I looked into this a bit and it looks like the root cause of the stale path is coming from onMouseOut in Boxes.jsx. The fix here seems minimal:

```js
// Boxes.jsx — onMouseOut
window.clearTimeout(timeout); 	

if (!event.relatedTarget || 
    (!isDescendantOrSelf(...) && !event.target.closest('#menuHolder'))
) {
    clearHovered();
    hoverProviderRef.current?.setHoveredPath(null);
}
```

We just need to make sure that 1) we always clear the timeout on the mouse out so that we don't end up with a stale path 2) also handle the missing case where relatedTarget is null (cursor left the iframe).

Once stale path is fixed, BoxesContextMenu can then just use currentPathRef.current as the targetPath (same ref pattern already used for selection/nodes) and then check for null. I tried this out with the test and seems to be working ok.

I think this could be a cleaner fix than working around the root issue — but lmk if there are stale scenarios I'm missing. Either way the elementsFromPoint approach works, just flagging this in case it simplifies things. lmk what you think with this approach.

gflores-jahia (commented):
added comment

src/javascript/JContent/EditFrame/Boxes/BoxesContextMenu.jsx
AKarmanov: This assumes that onMouseOut is called but it may not be. You can navigate away from the element via the menu itself (use the menu to move mouse out of the box) and then you still have the box selected with the path. In that case clicking elsewhere on the page will open another menu for that currently selected element which is not even hovered over. That's the edge case that this is trying to solve. You can see here that the menu is opened when the cursor is outside of the box but the element is still selected because the event is never fired (this has timeout removal as you suggested):
Screenshot 2026-06-24 at 7 17 58 AM

Relying on that event is a silpery slope.

sdusolle (approved):

src/javascript/JContent/EditFrame/Boxes/BoxesContextMenu.jsx
gflores-jahia: That's an interesting use case, thanks for pointing that out. But yeah it's a bit out of scope for the story so yeah we can go with your solution for now.

gflores-jahia (approved):

PR #2372, opened 2026-06-05
feat: translate layout in advanced editor

GauBen:
Working on the premises of  https://github.com/Jahia/content-versioning/issues/8: the translation layout is now part of the advanced edition layout

image

- The two columns were swapped
- Translate menu entry is now not displayed (instead of greyed) for single language site.
- Default language to translate from is now the first non matching with the current edit language.  

It should be somewhat easy to add new options (e.g. content versioning) next to edit/translate/advanced options

## Actions

- [x] Move the translation layout to the advanced editing layout
- [x] Remove unused translation code
- [x] Don't show the translation option when the site is in a single language
- [x] Update the tests

PR #2447, opened 2026-06-23
fix: language switch no longer remounts the content editor form

dgriffon:
## Summary

  Switching the editing language in the content editor previously unmounted and remounted the entire form, causing a full
  visual flash via `LoaderOverlay`. This fix keeps the form mounted throughout the language switch while preserving all
  unsaved edits in other languages.

  ## What changed

  **`useFormDefinitions.js`** — keeps a `previousData` ref of the last successfully adapted form definition. While the new
  language's data is loading (`loading: true`), returns the stale data with `loading: false, isRefetching: true` instead
  of propagating `loading: true`. The initial load (no previous data yet) still shows the overlay as before.

  **`ContentEditor.context.js`** — stores the last fully-valid `editorContext` and `sections` in refs. While `loading`,
  `siteInfoResult.loading`, `!ranAllHooks`, or `isRefetching` is true, renders children with the stale context instead of
  `<LoaderOverlay/>`. The `isRefetching` flag is key: it ensures the context does not transition to `lang=newLang`
  prematurely while `initialValues` are still stale — `lang` and `initialValues` always transition atomically when fresh
  data arrives, which is what `I18nContextHandler` relies on to apply saved per-language edits against the correct Formik
  base values.

  **`Edit.jsx`** — adds `enableReinitialize` to `Formik` so the form picks up the new language's `initialValues` when
  fresh data arrives, without an unmount/remount cycle. The existing `I18nContextHandler` then correctly restores any
  previously saved edits for the new language on top of those server values.

  ## Behaviour preserved

  - Initial load still shows the `LoaderOverlay` (no previous data in the ref)
  - Unsaved edits in the previous language are saved to `i18nContext` by `useSwitchLanguage` before the switch, and
  restored by `I18nContextHandler` after the new language loads — unchanged
  - `isDirty` / unsaved-changes guard remains correct throughout the transition
  - Save flow (`reFetchObservableQueries` after save) is unaffected: `resetI18nContext` clears the context before the
  refetch, and `enableReinitialize` correctly resets to the freshly saved server values

### DIFF
--- a/.chachalog/pi1abfXf.md
+++ b/.chachalog/pi1abfXf.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": patch
+---
+
+Switching the editing language no longer reloads the content editor
--- a/src/javascript/ContentEditor/ContentEditor/Edit.jsx
+++ b/src/javascript/ContentEditor/ContentEditor/Edit.jsx
@@ -66,6 +66,7 @@ export const Edit = () => {
         <>
             <PublicationInfoContextProvider uuid={nodeData.uuid} lang={lang}>
                 <Formik
+                    enableReinitialize
                     validateOnMount
                     validateOnChange={false}
                     initialValues={initialValues}
--- a/src/javascript/ContentEditor/ContentEditor/Edit.jsx
+++ b/src/javascript/ContentEditor/ContentEditor/Edit.jsx
@@ -62,12 +62,9 @@ export const Edit = () => {
         });
     }, [client, t, notificationContext, editCallback, contentEditorConfigContext, lang, nodeData, sections, i18nContext]);
 
-    // With enableReinitialize, Formik resets values whenever the initialValues REFERENCE changes.
-    // The adapter can recompute initialValues (same content, new identity) on incidental re-renders;
-    // letting those through resets in-progress edits — e.g. a just-toggled mixin fieldset collapses
-    // mid-interaction (mixins.cy.ts race). Only hand Formik a new reference when it may legitimately
-    // differ: language switch or a different node. Same-lang/same-node keeps pre-#2447 semantics
-    // (never reinitialize); imperative resetForm calls (save) are unaffected.
+    // Formik reinitializes whenever the initialValues REFERENCE changes, and the adapter recomputes
+    // it (same content, new identity) on incidental re-renders — which would reset in-progress
+    // edits. Only hand Formik a new reference on a genuine language or node change.
     const stableInitialValuesRef = useRef({lang, uuid: nodeData.uuid, initialValues});
     if (stableInitialValuesRef.current.lang !== lang || stableInitialValuesRef.current.uuid !== nodeData.uuid) {
         stableInitialValuesRef.current = {lang, uuid: nodeData.uuid, initialValues};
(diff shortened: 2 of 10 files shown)


sdusolle (dismissed):

sdusolle (dismissed):

eallais:
Hello :slightly_smiling_face:
Yes add it to the board as it could be a side effet of https://github.com/Jahia/jcontent/commit/d6bda91a6d1820cf63abfd11cdf3fee46a5ecd6d
It is we'll see if we want to share that with GOAT or fix it ourselves

sdusolle (dismissed):

.chachalog/pi1abfXf.md
```diff
@@ -0,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": patch
+---
+
+Switching the editing language no longer reloads the content editor
```
gflores-jahia, line 6: As it's follow up to PR https://github.com/Jahia/jcontent/pull/2372, we should just update that changelog (or keep as-is) instead of adding a new entry

eallais:
## Branch rebased + integration failures fixed

**Heads-up: the branch was rebased onto `main` and force-pushed** (history rewritten). Previous approval will need to be re-given.

### Why the rebase
The branch predated #2491 (accordion customization), whose `addContextMenuTargetToActions` export other modules now expect from jcontent's federation shared scope — builds of this branch crashed the app (`TypeError: addContextMenuTargetToActions is not a function`) against a current module ecosystem. Rebase was clean, no conflicts.

### Fix 1 — `uploadMedia.cy.ts` flaky (`test: wait for marked-for-deletion state…`)
`File.markForDeletion()` returned as soon as the mark dialog was dismissed, so `deletePermanently()` could reopen the context menu before `jmix:markedForDeletion` propagated to the grid card — the `deletePermanently` action was then missing (20s timeout). Now waits for the card status, same guard `Folder.markForDeletion` already uses. Covers all 5 upload tests.

### Fix 2 — `mixins.cy.ts` regression (`fix: stop Formik reinitialize and stale-serving…`)
Root cause found by reproducing locally against a full stack: this PR adds `enableReinitialize` to the Formik in `Edit.jsx`. Formik resets values whenever the `initialValues` **reference** changes — and the adapter recomputes `initialValues` (same content, new identity) on incidental re-renders. Any such re-render landing mid-interaction reset in-progress edits: a just-toggled mixin fieldset collapsed (menu items vanished under `cy.click()`, fields flipped back to disabled under `cy.type()`). Three coupled changes:

1. **`Edit.jsx`** — hand Formik a new `initialValues` reference only when `lang` or the node changes; same-lang/same-node keeps pre-PR never-reinitialize semantics (imperative `resetForm` on save unaffected).
2. **`useFormDefinitions.js`** — `isRefetching` (stale-data serving) only when `lang` actually moved away from the cached data's lang; structural refetches (save, mixin toggle, dependent field) keep the original loading behaviour.
3. **`ContentEditor.context.js`** — stale context + refetch-blocker overlay only when `isRefetching`; non-language refetches fall back to the `LoaderOverlay` exactly as before this PR.

### Local validation (full Jahia stack, 8 runs each)
| Build | `should add mixins to a page` | `translate.cy.ts` (this PR's lang-switch guards) |
|---|---|---|
| main snapshot (baseline) | 8/8 | — |
| this branch without fix | 2/5 | — |
| this branch + fix | **8/8** | **8/8** |

The language-switch no-remount goal of this PR is preserved (all 8 `translate.cy.ts` tests pass, incl. the 3 mode-persistence regression guards).

sdusolle (approved):

gflores-jahia (approved):

PR #2472, opened 2026-06-25
fix: preview width in Firefox

### LINKED ISSUE #2218: Use the same code for content preview in jcontent and content editor
**The goal of the spike is to** 
Use the same code to manage previews in jContent and in Content Editor.
We want to have one code only to support and make maintenance easier, as well as better results displayed to editors.
Also, we want this unique preview code because we may want to reuse it later when a user add a new content in a page, by providing a preview of a content-type and thius make things easier because more "visual"

### Outcomes
First implementation + a story to "finsih" properly the implementation if there are some defects or problems  remaining.

### Notes

Useful information (Technical, Design)

gflores-jahia:
### Description

Fix issue with current side panel styling not working in Firefox:

image

### DIFF
--- a/src/javascript/JContent/SidePanel/SidePanel.scss
+++ b/src/javascript/JContent/SidePanel/SidePanel.scss
@@ -1,5 +1,5 @@
 .root {
-    width: stretch;
+    width: 100%;
     height: 100%;
 }
 


AKarmanov (approved):

gflores-jahia:
I looked into the failing tests and not related to preview at all. Will verify if still on merge to main workflow

PR #2479, opened 2026-06-26
feat: support node type resolver in form generation

baptistegrimaud:
Closes https://github.com/Jahia/content-versioning/issues/76.

### Description

Adds `NodeTypeResolver` interface to decouple form structure from JCR node state, enabling external modules to control mixin activation and type checks. Useful for generating forms for historical snapshots or other custom contexts.

> [!TIP]

### DIFF
--- a/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
+++ b/src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
@@ -106,16 +106,25 @@ public class EditorFormServiceImpl implements EditorFormService {
         }
     }
 
+
+
     private Form getEditorForm(ExtendedNodeType primaryNodeType, JCRNodeWrapper existingNode, JCRNodeWrapper parentNode, Locale uiLocale, Locale locale) throws EditorFormException {
+        return getEditorForm(primaryNodeType, existingNode, parentNode,
+            existingNode != null ? new JcrNodeTypeResolver(existingNode) : null,
+            uiLocale, locale);
+    }
+
+    @Override
+    public Form getEditorForm(ExtendedNodeType primaryNodeType, JCRNodeWrapper existingNode, JCRNodeWrapper parentNode, NodeTypeResolver nodeTypeResolver, Locale uiLocale, Locale locale) throws EditorFormException {
         final String mode = existingNode == null ? CREATE : EDIT;
         final JCRNodeWrapper currentNode = EDIT.equals(mode) ? existingNode : parentNode;
 
         try {
             final JCRSiteNode site = currentNode.getResolveSite();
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
```diff
@@ -130,9 +139,9 @@ private Form getEditorForm(ExtendedNodeType primaryNodeType, JCRNodeWrapper exis
                 addFormNodeType(extendMixin, site, mergeSet, locale, true, processedNodeTypes, nodeTypes);
             }
 
-            // Mixins added on node
-            if (existingNode != null) {
-                for (ExtendedNodeType mixinNodeType : existingNode.getMixinNodeTypes()) {
+            // Mixins added on node (from the node type resolver)
+            if (nodeTypeResolver != null) {
```
gflores-jahia, line 143: There's a null check here but not on the other usages. 

We could do maybe two options - either add the null checks for each resolver usage, or do a null check in the beginning and add a default JCR implementation if null. wdyt ?

gflores-jahia (commented):
Added a comment about null checks

src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
baptistegrimaud: Good point, I missed it, thanks!
Because `getEditorForm(ExtendedNodeType primaryNodeType, JCRNodeWrapper existingNode, JCRNodeWrapper parentNode, Locale uiLocale, Locale locale)` calls `getEditorForm(ExtendedNodeType primaryNodeType, JCRNodeWrapper existingNode, JCRNodeWrapper parentNode, NodeTypeResolver nodeTypeResolver, Locale uiLocale, Locale locale)` with a `null` `nodeTypeResolver` parameter when there is no `existingNode`, I added a `null` check when using the resolver.
Addressed in [`795a09f` (this PR)](https://github.com/Jahia/jcontent/pull/2479/changes/795a09ff73966c8a1c095575c7ae1edc4c073be7)

baptistegrimaud: FYI, I also added a null check in the constructor of `JcrNodeTypeResolver` via [`ab5a692` (this PR)](https://github.com/Jahia/jcontent/pull/2479/changes/ab5a6920dcdd3419eb222fb8d96b7e1ff7ec0d55)

gflores-jahia (approved):

src/main/java/org/jahia/modules/contenteditor/api/forms/EditorFormServiceImpl.java
gflores-jahia: all good, thanks !

PR #2440, opened 2026-06-22
feat: Refactor visibility in react wrap-up

cedmail:
### Description
Remove logout in afterEach()

#### Source code
- [ ] I've shared and documented any breaking change
- [ ] I've reviewed and updated the jahia-depends

#### Tests
- [ ] I've provided Unit and/or Integration Tests
- [ ] I've updated the parent issue with required manual validations

> [!TIP]
> Documentation to guide the reviews: [How to do a code review](https://jahia-confluence.atlassian.net/wiki/spaces/PR/pages/2064660/How+to+do+a+code+review+-+Ref+ISSOP08.A14006)

PR #2491, opened 2026-06-30
feat: Accordion customization options for context menu, columns

### LINKED ISSUE #2480: Jcontent improvements for accordion customizations

The archiving module revealed some blockages in jContent (lack of flexibility) and at the same time is highly requested by some customers. We are blocked since 4 or 5 monthes.

### Story

As a module developer registering a custom accordion item in jContent,
I want to configure the content table's columns, right-click context menu, and header status badge through `tableConfig`,
so that the table UI is appropriate for my content type without forking jContent internals.

## Testable scenarios

**Setup**
Given a module that registers a custom accordion item with a `tableConfig` that sets `contextualMenu`, `header.showStatus: false`, and a `columns` list
And the module is enabled on a site

**Case 1 – Custom columns**
When the user navigates to the custom accordion in list mode
Then only the columns declared in `tableConfig.columns` are visible in the table header

**Case 2 – Custom context menu**
When the user right-clicks a content row
Then only the actions registered for the custom menu target are shown
And standard actions not in that set (e.g. Delete, Copy) are absent

**Case 3 – Hidden status badge**
When the user navigates to the custom accordion
Then no badges are rendered in the content header (bottom right)

**Case 4 – Default accordion unaffected**
When the user navigates to a standard accordion (e.g. pages, content-folders)
Then all default columns, the standard context menu, and the status badge appear as normal

### Test Strategy

Cypress tests in `shard2/accordionConfig.cy.ts` cover all four cases using a dedicated `tableconfig-test` accordion registered by `jcontent-test-module`.

### Tech Notes

- New public API: `addContextMenuTargetToActions(targetName, actionKeys[])` — assigns a menu target to existing registry actions so they appear in a custom context menu. Exported from `@jahia/jcontent`.
- New public API: `columnDefinitions` namespace — exports all built-in column definition objects (`name`, `type`, `lastModified`, etc.) for use in `tableConfig.columns`. Exported from `@jahia/jcontent`.
- `tableConfig.columns` accepts either string IDs (resolved against `allColumnData`) or inline react-table column definition objects, allowing fully custom columns alongside built-ins.
- Media accordion's column set (`mediaColumnData`) is now declared in its own `tableConfig.columns` rather than as a hardcoded branch in `ContentTable`, making it consistent with the new pattern.

### Labels

En: Accordion table customization (columns, context menu, header status)
Fr: Personnalisation du tableau d'accordéon (colonnes, menu contextuel, statut en-tête)
De: Accordion-Tabelle anpassen (Spalten, Kontextmenü, Header-Status)

### Original spike goals

**The goal of the spike is to** 
- see how a module could hide all badges in jcontent header (could be based on the accordeon, the node type...)
- see how a module can specify the desired columns
- see how a module could register a new entry point for a custom menu or allow to define a custom menu based on a node-type

### Outcomes
- POC if possible
- Evaluation of the effort to implement those changes.

gflores-jahia:
### Description

Added new tableConfig options for accordion items

contextualMenu to override the right-click menu
header.showStatus to hide the publication-status badge
columns to control which table columns are shown.

#### Test

Added new test suite `accordionConfig` to cover each scenario

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/tests/assets/provisioning.yml
+++ b/tests/assets/provisioning.yml
@@ -7,7 +7,6 @@
 
 # This modules were added during the switch from using jahia-discovery to jahia-ee
 - installBundle:
-    - 'mvn:org.jahia.modules/legacy-default-components'
     - 'mvn:org.jahia.modules/dx-base-demo-core'
     - 'mvn:org.jahia.modules/dx-base-demo-components'
     - 'mvn:org.jahia.modules/dx-base-demo-templates'
--- a/.chachalog/HKqkVPCe.md
+++ b/.chachalog/HKqkVPCe.md
@@ -1,0 +1,6 @@
+---
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": minor
+---
+
+Added new `tableConfig` options for accordion items - `contextualMenu` to override the right-click menu, `header.showStatus` to hide the publication-status badge, and `columns` to control which table columns are shown. (#2491)
--- a/src/javascript/JContent.assignActionAndMenuTargets.js
+++ b/src/javascript/JContent.assignActionAndMenuTargets.js
@@ -39,6 +39,10 @@ const assignTargetsForActions = (targetActions, registry) => {
     });
 };
 
+export const addContextMenuTargetToActions = (targetName, actions) => {
+    assignTargetsForActions({[targetName]: actions}, registry);
+};
+
 const actionTargetAssignments = {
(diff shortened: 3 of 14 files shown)


gflores-jahia:
Failure still coming from visibility tests, to be addressed in PR https://github.com/Jahia/jcontent/pull/2440

tests/assets/provisioning.yml
```diff
@@ -7,7 +7,6 @@
 
 # This modules were added during the switch from using jahia-discovery to jahia-ee
 - installBundle:
-    - 'mvn:org.jahia.modules/legacy-default-components'
```
gflores-jahia, line 10: Already added as part of core-modules https://github.com/Jahia/jahia-pack-private/blob/main/core-modules/pom.xml#L346. No need to install separately

sdusolle (approved):

r3dm1ke (approved):

PR #2501, opened 2026-07-01
feat: export editor-form API packages for external modules

eallais:
## What

Export two Content Editor API packages from jContent so external modules can consume them at runtime:

- `org.jahia.modules.contenteditor.api.forms` — `EditorFormService`, `NodeTypeResolver`, `EditorFormException`
- `org.jahia.modules.contenteditor.api.forms.model` — `Form`

Only `org.jahia.modules.contenteditor.graphql.api.forms` was exported before, so these types could be compiled against but **not OSGi-wired** from another bundle.

## Why

#2479 introduced `NodeTypeResolver` as an extension point explicitly *"for external modules [to] supply alternative implementations … via `EditorFormService.getEditorForm(...)`"*. But the packages holding that interface (and the `Form` return type) aren't exported, so a separate bundle cannot import them — it stays `INSTALLED` / never `ACTIVE` at deploy.

Concretely this blocks **Jahia/content-versioning#135**: a `versionHistoryEditorForm` GraphQL field that builds a Content Editor form for a historical version via a `SnapshotNodeTypeResolver implements NodeTypeResolver`, delegating to `EditorFormService`. It compiles, but the content-versioning bundle can't resolve until these packages are exported.

## Change

Two lines added to `<export-package>` — no code change.

Refs #2479
Unblocks Jahia/content-versioning#135

### DIFF
--- a/.chachalog/Zx7Ke2Qp.md
+++ b/.chachalog/Zx7Ke2Qp.md
@@ -1,6 +1,0 @@
----
-# Allowed version bumps: patch, minor, major
-"@jahia/jcontent": minor
----
-
-Export the editor-form API packages (`org.jahia.modules.contenteditor.api.forms` and `org.jahia.modules.contenteditor.api.forms.model`) so external modules can implement the `NodeTypeResolver` extension point (#2479) and generate editor forms through `EditorFormService`. Previously only `graphql.api.forms` was exported, so these types were unreachable across bundles. (#2479)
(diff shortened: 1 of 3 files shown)


.chachalog/Zx7Ke2Qp.md
gflores-jahia, line 1: Since this is a follow-up to a previous PR, https://github.com/Jahia/jcontent/pull/2479/changes. I'd recommend just merging this and adding PR links to the original `.chachalog/nodeTypeResolver.md` changelog for both #2479 and this one.

gflores-jahia (commented):
@eallais added changelog comment

.chachalog/Zx7Ke2Qp.md
eallais: Done :)

gflores-jahia (approved):

PR #2525, opened 2026-07-07
feat: add workflows shortcut to modes dropdown

### LINKED ISSUE #2458: Access to Workflow setting
In the new dropdown menu (Modes), in the section "Advanced options"

add an entry "Workflows"
the entry opens directly the gwt modal
beware that we must keep the check that existed in the advanced options panel and verify if there are unsaved changes before showing the GWT interface (reuse exactly the same logic and ui)

r3dm1ke:
### Description
This PR adds a workflows shortcut to the modes dropdown in CE.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
+++ b/src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import {DisplayAction, DisplayActions, registry} from '@jahia/ui-extender';
 import {ButtonRendererShortLabel, getButtonRenderer} from '~/ContentEditor/utils';
 import {truncate} from '~/utils';
-import {ButtonGroup, Dropdown, Header, Separator} from '@jahia/moonstone';
+import {ButtonGroup, Dropdown, Header, Separator, Workflow} from '@jahia/moonstone';
 import styles from './EditPanelHeader.scss';
 import {PublishMenu} from './PublishMenu';
 import {useTranslation} from 'react-i18next';
@@ -15,6 +15,8 @@ import {HeaderButtonActions, HeaderThreeDotsActions} from '../HeaderActions';
 import {ContentTypeChip} from '../ContentTypeChip';
 import {useNodeChecks} from '@jahia/data-helper';
 import {Constants} from '~/ContentEditor/ContentEditor.constants';
+import {useEngineTabAvailability} from '~/ContentEditor/editorTabs/AdvancedOptions/AdvancedOptionsNavigation/useEngineTabAvailability';
+import {useOpenEngineTabsWithConfirmation} from '~/ContentEditor/editorTabs/AdvancedOptions/AdvancedOptionsNavigation/useOpenEngineTabsWithConfirmation';
 
 const ButtonRenderer = getButtonRenderer({
     defaultButtonProps: {size: 'big', color: 'accent'}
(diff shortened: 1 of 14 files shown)


tests/cypress/page-object/contentEditor.ts
```diff
@@ -338,4 +338,9 @@ export class ContentEditor extends BasePage {
         this.switchToAdvancedMode();
         return this.switchToAdvancedOptions();
     }
+
+    openWorkflowsFromModesDropdown() {
```
gflores-jahia, line 342: Could be more reusable if we make this more generic e.g. `selectModesDropdown(roleSelector)` and pass in role selector for the modes dropdown instead.

I'll let you decide if we do it in this PR or follow-up ones.

src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
```diff
@@ -43,6 +45,32 @@ export const EditPanelHeader = ({
...
+
+    // The moonstone Dropdown supports either a flat list or groups only, so entries switch
+    // to grouped form (with an unlabeled group on top) once the advanced options section shows
+    const dropdownData = hasWorkflowsEntry ? [
+        {groupLabel: '', options: tabOptions},
+        {
+            groupLabel: t('label.contentEditor.edit.tab.advanced'),
+            options: [{
+                value: 'workflows',
+                label: t('label.contentEditor.edit.tab.workflows'),
```
AKarmanov, line 67: The title should be available on the workflow tab from the API call: {
    "id": "workflow",
    "title": "Workflow",
    "requiredPermission": "viewWorkflowTab"
}

src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
```diff
@@ -43,6 +45,32 @@ export const EditPanelHeader = ({
...
+    const tabOptions = tabs.map(tab => ({
+        value: tab.value,
+        label: t(tab.buttonLabel),
+        iconStart: tab.buttonIcon,
+        attributes: {'data-sel-role': tab.dataSelRole}
+    }));
+
+    // The moonstone Dropdown supports either a flat list or groups only, so entries switch
+    // to grouped form (with an unlabeled group on top) once the advanced options section shows
+    const dropdownData = hasWorkflowsEntry ? [
```
AKarmanov, line 61: The user may not have workflow but have tab items, so this will have to be adjusted accordingly as we add more items (note to self).

src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
```diff
@@ -43,6 +45,32 @@ export const EditPanelHeader = ({
         .find({target: 'editHeaderTabsActions'})
         .filter(tab => tab.isDisplayable(ctx) && (!tab.requiresAdvancedPermission || res.checksResult));
 
+    const {availableTabs: workflowTabs} = useEngineTabAvailability(['workflow']);
+    const {openTabs: openWorkflows, confirmationDialog: workflowsConfirmationDialog} = useOpenEngineTabsWithConfirmation(['workflow']);
```
AKarmanov, line 49: Would I have to create another instance of useOpenEngineTabsWithConfirmation(['workflow']) but with `liveroles` to enable Live roles.

src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
AKarmanov: I wonder if we can use that somehow and also use that info to maybe streamline the whole thing and add live/edit roles to is as well?

src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
gflores-jahia: I think it can be possible to make this generic, i.e. being able to pass in the tab name in runtime when it's called from https://github.com/Jahia/jcontent/pull/2525/changes#diff-dded2b2a22cf72f2665c5d0dcbea116db9626744a655630fab40f5f2389a0afaR142 (could be done as a follow up with the role tabs)

src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
gflores-jahia: Rename would be nice too e.g. `hasGwtEntries` or similar

src/javascript/ContentEditor/ContentEditor/EditPanel/EditPanelHeader/EditPanelHeader.jsx
gflores-jahia: wondering if this can be moved to `useEngineTabAvailability` as part of `availableTabs` object as well, and iterate through availableTabs options.

gflores-jahia (dismissed):
Added suggestions, but can be done as a follow up to the rest of GWT actions.

AKarmanov (approved):

PR #2531, opened 2026-07-08
fix: Replace message rendering in Delete dialog

gflores-jahia:
### Description

Remove `innerhtml` raw rendering when displaying delete message. 

Replace with use of `<Trans>` component from i18n package. Refactored `renderLabel` as a separate file.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/main/resources/javascript/locales/de.json
+++ b/src/main/resources/javascript/locales/de.json
@@ -477,7 +477,7 @@
         },
         "mark": {
           "title": "Zur Löschung markiert",
-          "item": "Sie sind dabei, <span>{{name}}</span> zu löschen.",
+          "item": "Sie sind dabei, <0>{{name}}</0> zu löschen.",
           "items": "Sie sind dabei, <span>{{count}} Element(e)</span> zu löschen, darunter <span>{{pages}} Seite(n)</span>.",
           "filesAndFolders": "Sie sind dabei, <span>{{count}} Element(e)</span> zu löschen, darunter <span>{{folders}} Ordner</span>.",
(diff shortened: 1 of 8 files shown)


sdusolle (approved):

src/main/resources/javascript/locales/de.json
```diff
@@ -480,23 +477,23 @@
         },
         "mark": {
           "title": "Zur Löschung markiert",
-          "item": "Sie sind dabei, <span>{{name}}</span> zu löschen.",
+          "item": "Sie sind dabei, <0>{{name}}</0> zu löschen.",
```
AKarmanov, line 480: So just curious, we no longer want to use span/strong etc. in translations or is it only for specific parts?

AKarmanov (approved):

src/main/resources/javascript/locales/de.json
gflores-jahia: We can still do this (we still use \<span\> and \<strong\> here) but is now parameterized. The indices in the translation json (<0>, <1>, etc). would be replaced by the components declared in order like in this example:

```
<Trans i18nKey={`jcontent:label.contentManager.deleteAction.${dialogType}.item`}>
    <span>{{name: firstNode && getName(firstNode)}}</span>
    <strong/>
</Trans>
```
so `<0>{{name}}</0>` will get rendered as `<span>{{firstNode && getName(firstNode)}}</span>`, `<1>` by `<strong>`, etc

https://react.i18next.com/latest/trans-component#alternative-usage-components-array

There were different options to specify this component replacement (some more straightforward) but I went with this one due to i18n version compatibility.

PR #2544, opened 2026-07-09
fix: Maintain button state as formik validation runs

AKarmanov:
### Description
Maintain button state as formik validation runs. This fixes the issue when two clicks are required to save the form if it is opened with a validation error, which is then fixed and save button is clicked without blur listener getting called.

#### Source code

#### Tests

> [!TIP]


src/javascript/ContentEditor/utils/getButtonRenderer.jsx
```diff
@@ -40,16 +41,18 @@ export const getButtonRenderer = ({labelStyle, defaultButtonProps, noIcon} = {})
             />
         );
```
gflores-jahia, line 42: [nitpick] not really necessary anymore. can be added inline on L52.

gflores-jahia (approved):

sdusolle (approved):

PR #2545, opened 2026-07-09
fix: Add graphql-dxm-provider version check in tag manager

gflores-jahia:
## Description 

- Revert updated jahia-depends for gql-dxm-provider
- Add graphql-dxm-provider module version check and register tag manager only if >= 3.8
  - workaround: Add current active version of dxm-provider module in jcontent.jsp

#### Misc

- Cleanup of unused jahia version modules

No automated tests as this requires installation of multiple dxm-provider modules and can cause instability with the tests with using multiple versions for a very specific usage in tag manager only. Require manual test instead.

#### Source code

#### Tests

> [!TIP]


smonier (changes requested):
Tested end to end on a fresh Jahia 8.2.3.2 instance (digitall site), swapping the active graphql-dxm-provider between 3.7.0 and 3.11.0-SNAPSHOT with this branch's jcontent build:

- `jcontent.jsp` correctly emits the active provider version (`"3.7.0"` / `"3.11.0-SNAPSHOT"`), and `register.js` hides/shows the Tags route accordingly. The mechanism works — thanks for this.

Two things before merge:

1. **Threshold**: `REQUIRED_GQL_DXM_VERSION = '>=3.8'` should be `'>=3.11'`. The `admin.jahia.tagManager` API was introduced by graphql-core #597, which ships in **3.11.0** (`main` is 3.11.0-SNAPSHOT). 3.8, 3.9 and 3.10 are released versions without that field, so with `>=3.8` they pass the gate and the Tag Manager page fails with `Validation error (FieldUndefined@[admin/jahia/tagManager])` — exactly the situation this PR is meant to prevent (reproduced against 3.7.0, same schema for 3.8–3.10).

2. **Rebase**: `smo-tagManager` was rebased onto `main` today, so the branch is now conflicting; a rebase onto the current base is needed.

Everything else looks good to me: `jahia-depends=3.5` is fine now that compatibility is enforced at runtime, `normalize()`/`satisfies()` handle `-SNAPSHOT` correctly, and the removal of the dead `ZipActionComponent`/`dxVersion` checks is correct. The Sonar findings are on the base branch's commits, not on this diff.

smonier:
Rebased `add-gql-version-check` onto the current `smo-tagManager` (`908b8c07`) — clean replay of your two commits, no content changes. Only the threshold change from the review is left.

smonier:
Pushed the two one-line fixes from the review directly to the branch (`register.js`): threshold `>=3.11`, and `globalThis` instead of `window` for Sonar S7764. Added the `🗒️ no-changelog` label: the customer-facing note for this requirement is in the Tag Manager changelog entry on `smo-tagManager` (`.chachalog/cc797195.md`), so no separate fragment is needed here.

smonier:
Correction to my earlier review: I asked for the threshold to be `>=3.11`, based on `graphql-dxm-provider` `main` currently being `3.11.0-SNAPSHOT` — I never actually checked when the `JahiaTagManager`/`admin.jahia.tagManager` API was **released**, only what `main` is at right now. It was verified (and just confirmed live against 3.9.1) that the API has been present in every release since **3.8.0** (cut 2026-07-27, right after #597 merged): 3.8.0, 3.8.1, 3.9.0, 3.10.0 all have it. Your original `>=3.8` was correct.

Reverted on `smo-tagManager`: `9c3165bd` (register.js threshold) and the changelog entry. Apologies for the churn — thank you for catching it.

PR #2561, opened 2026-07-15
feat(image-editor): replace built-in image editor with Filerobot-based editor

romain-pm:
> **Draft / WIP** — opening for early review.

## What
Replaces jContent's built-in image editor (the custom crop / resize / rotate dialog) with an editor based on [Filerobot Image Editor](https://github.com/scaleflex/filerobot-image-editor) (`react-filerobot-image-editor` v4, MIT), giving editors crop, rotate, flip, annotate, filters, finetune and resize in one component.

## Why
The built-in editor only supported basic crop/resize/rotate and was a fair amount of bespoke UI to maintain. Filerobot provides a far richer, well-maintained editing experience out of the box.

## Changes
- **Add** `src/javascript/JContent/actions/editImage/FilerobotEditor/` — `FilerobotEditor.jsx` (lazy-loaded), `filerobotTheme.js` (moonstone-aligned theme), `useSaveEditedImage.js`, styles.
- **Rewire** `editImageAction.jsx` to open the new editor.
- **Remove** the old `ImageEditorDialog` and its `ResizePanel` / `RotatePanel` / `ImageEditorPreview` / `UnsavedChangesDialog` sub-components and tests.
- **Deps**: `react-filerobot-image-editor`, `react-konva`, `konva` (pinned to Filerobot's version), `styled-components`; `webpack.config.js` plain-CSS rule widened for the editor's styles.
- **i18n**: en/fr/de updated.
- **Tests**: Cypress e2e `tests/cypress/e2e/jcontent/imageEditor.cy.ts` + `page-object/imageEditor.ts`.

## Notes / open points for reviewers
- React compatibility: pinned `react-filerobot-image-editor` v4 (peer `react >=17`) to match jContent's React 18; `react-konva` 18.x / `konva` 9.3.6 to avoid a duplicate React. v5 requires React 19.
- Save path / output format and the exact toolbar set are still being tuned — feedback welcome.
- Verify bundle size impact (Filerobot + konva) is acceptable; it is lazy-loaded so it should not affect initial jContent load.


gflores-jahia (commented):
Very early feedback and didn't go into details yet. So far the setup looks good. Found some weird case when editing realtor images on luxe where dimensions looks to be unknown and defaults to 300x300 and crops automatically.

UI also still needs to be updated on the reset, and need to keep track of dirty editor, and maybe show a confirm dialog on exit.

Preview icon on the copy also doesn't seem to be generating. Not sure if it's a preview-related bug instead, or that save is still WIP.

One thing I did notice with webpack config is the change from allowlist to a catch-all exclude and can capture all in node_modules dependency. This might cause some css loading conflicts and possibly bleed into the UI as it's a shared module federation remote. 

Maybe we can try to nail down exact required css and keep an explicit include list of the know filerobot-related packages rather than switch to exclude-all.

Lastly, noticed branch is a bit out of date and does not include the major css module refactor. Might be worth it to update the branch in case there's any side effects.

PR #2553, opened 2026-07-13
fix: Add actions to menu

AKarmanov:
### Description
Add export/import actions to content tree menu.

#### Source code
- [ ] I've shared and documented any breaking change
- [ ] I've reviewed and updated the jahia-depends

#### Tests
- [ ] I've provided Unit and/or Integration Tests
- [ ] I've updated the parent issue with required manual validations

> [!TIP]
> Documentation to guide the reviews: [How to do a code review](https://jahia-confluence.atlassian.net/wiki/spaces/PR/pages/2064660/How+to+do+a+code+review+-+Ref+ISSOP08.A14006)

PR #2565, opened 2026-07-15
test: cover accordionContentActions export/import entries (#2471)

### LINKED ISSUE #2471: PageBuilder: NavMenuText cannot be exported

8.2.3.0, jcontent 3.6.1

### Steps to reproduce

Create a NavMenuText entry

### Current Behavior

From Page Builder in the context menu, you don't have a possibility to export the NavMenuText:

Image

### Desired Behavior

Because this was possible with Page Composer, it should be possible with Page Builder too. 

It definitely makes sense to export a NavMenuText and subpages, because you create the navMenuText to logical collect pages (and this could be exported, to import on another installation). 

Workaround: Use the list view to export the NavMenuText

danielaebenberger:
### Description
Unit test for the config change in #2553 (fixes #2471): `accordionContentActions` was missing
`export`/`exportPage`/`import`, so Page Builder's accordion tree menu didn't offer them for a
`navMenuText` node. Adds a direct assertion on the array so this can't silently regress.

`actionTargetAssignments` is exported (was module-private) so it's directly testable.

#### Source code
- [x] No breaking change
#### Tests
- [x] Unit test added

### DIFF
--- a/src/javascript/JContent.assignActionAndMenuTargets.js
+++ b/src/javascript/JContent.assignActionAndMenuTargets.js
@@ -43,7 +43,7 @@ export const addContextMenuTargetToActions = (targetName, actions) => {
     assignTargetsForActions({[targetName]: actions}, registry);
 };
 
-const actionTargetAssignments = {
+export const actionTargetAssignments = {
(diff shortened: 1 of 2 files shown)


danielaebenberger:
Closing per review feedback: this array only regresses via a deliberate edit, and the accordion menu's existing permission-aware Cypress check (menuActions/visibility.cy.ts, being expanded in #2566) is a strictly stronger guard than this static array assertion. Not worth maintaining both.

PR #2566, opened 2026-07-15
test: expand accordion action-visibility for export/import (#2471)

### LINKED ISSUE #2471: PageBuilder: NavMenuText cannot be exported

danielaebenberger:
### Description
Reworked per review feedback. Rather than a new spec asserting the shared `ExportDialog`
component opens (already covered by `menuActions/export.cy.ts`/`import.cy.ts`, and the same
component behind every menu that offers export), this extends the existing, permission-aware
`'Displays accordionContent actions'` test in `menuActions/visibility.cy.ts`:

- The `'home'` page node case now also asserts `Export` and `Import content` — the two entries
  #2553 added to `accordionContentActions` for page-type nodes (fixes #2471).
- A new case asserts `Export` on a `jnt:navMenuText` node — the ticket's own reported scenario,
  gated by a different action (`export`, not `exportPage`, since `navMenuText` extends
  `jnt:content` rather than `jnt:page`).

Both stop at menu-item presence (already permission-resolved by this test's existing pattern),
not at the export mechanism itself.

#### Source code
- [x] No breaking change
#### Tests
- [x] E2e test added


AKarmanov (approved):
Thanks for the PR!

sdusolle (approved):

PR #2569, opened 2026-07-16
fix: apply textarea height from the field definition (#2437)

AKarmanov:
### Description
Allow to set textarea height if it is available in selectorOptions. The height will be set as max height and it will not be possible to make it larger with UI.
#### Source code

#### Tests

> [!TIP]


gflores-jahia (approved):

gflores-jahia (commented):
don't forget to link PR to story ;)

PR #2574, opened 2026-07-17
refactor: extract editForm sections into a reusable GraphQL fragment

### LINKED ISSUE #2554: Export the editForm sections selection as a reusable GraphQL fragment
Content-versioning displays previous content versions using content editor, and therefore contains a big copy/paste of the content editor gql query:

https://github.com/Jahia/content-versioning/blob/b745e17b9dae8fb0196edb24278762c67c513b82/src/javascript/components/VersionSnapshotForm/snapshotForm.gql-queries.ts#L50-L66

The goal of this ticket is to expose a gql fragment to the federation instead, so that jcontent can evolve this query without consumers noticing

delallioui:
Closes #2554

Extract create/edit content editor query into a fragment, exposed it to the federation

The query used to be slightly different and they're now one and the same (because it worked), there is a follow-up ticket to see if it was the right call: https://github.com/Jahia/jcontent/issues/2584


gflores-jahia (commented):
code looks good. 

copilot suggests to narrow down the check for the fields (make sure coming from `form` fields and not elsewhere)

src/javascript/ContentEditor/ContentEditor/edit.gql-queries.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        it('spreads the sections fragment and bundles its definition', () => {
+            expect(printed).toContain('...EditFormSections');
+            expect(printed).toContain('fragment EditFormSections on GqlEditorForm');
+        });
+
+        it('keeps the top-level form fields selected directly', () => {
+            for (const field of ['name', 'displayName', 'description', 'hasPreview', 'showAdvancedMode']) {
+                expect(printed).toContain(field);
+            }
+        });
```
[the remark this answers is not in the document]
delallioui: confirmed. Fixed in 8589c9ba: the test now walks the `DocumentNode` AST (`forms → editForm` selection set) and asserts these fields are **direct `Field` selections** of `editForm`, plus a new assertion that `EditFormSections` is spread directly there. Dropping any top-level field now fails the test.

src/javascript/ContentEditor/ContentEditor/edit.gql-queries.spec.js
GauBen, line 1: Remove this test file, we are not unit testing anything useful here

src/javascript/ContentEditor/ContentEditor/edit.gql-queries.js
```diff
@@ -128,71 +128,81 @@ const NodeDataFragment = {
...
 };
 
-export const EditFormQuery = gql`
-    query editForm($uilang:String!, $language:String!, $uuid: String!, $writePermission: String!, $childrenFilterTypes: [String]!) {
-        forms {
-            editForm(uiLocale: $uilang, locale: $language, uuidOrPath: $uuid) {
+// The section → fieldSet → field selection shared by every editor-form query. Exported so other
+// consumers (e.g. the content-versioning version-history panel, which renders a read-only snapshot
+// form with the same shape) can spread it instead of copy-pasting the selection and drifting from it.
+export const EditFormSectionsFragment = gql`
```
GauBen: This thing needs to be made available to the federation (/src/javascript/shared/index.js)

src/javascript/ContentEditor/ContentEditor/edit.gql-queries.js
```diff
@@ -128,71 +128,81 @@ const NodeDataFragment = {
...
+
+export const EditFormQuery = gql`
+    query editForm($uilang:String!, $language:String!, $uuid: String!, $writePermission: String!, $childrenFilterTypes: [String]!) {
+        forms {
+            editForm(uiLocale: $uilang, locale: $language, uuidOrPath: $uuid) {
+                name
+                displayName
+                description
+                hasPreview
+                showAdvancedMode
```
GauBen: Not sure why these fields were left out, are they not needed to render content editor?

Do not just move them to the fragment, assess the necessity to have them in the fragment vs here

GauBen: Indeed in content-versioning we only use section, is this ok? https://github.com/Jahia/content-versioning/blob/b745e17b9dae8fb0196edb24278762c67c513b82/src/javascript/components/VersionSnapshotForm/snapshotForm.gql-queries.ts#L50

The fact that it works is not a proof that it's ok and that content editor is not falling back to a default value/global context/whatever

src/javascript/ContentEditor/ContentEditor/edit.gql-queries.js
delallioui: Good call — done in 141080fa

src/javascript/ContentEditor/ContentEditor/edit.gql-queries.spec.js
delallioui: Fair enough — on reflection you're right that this mostly restates the query's own selection rather than testing behaviour, and `adaptEditFormData` already covers the consumed shape. Removed in 0a81bc04.

src/javascript/ContentEditor/ContentEditor/edit.gql-queries.js
delallioui: Kept inline on purpose. EditFormQuery still selects them, so CE is unchanged, the fragment is scoped to sections because that's all consumers reuse, and hasPreview/showAdvancedMode are editor-only chrome the read-only version panel neither shows nor reads (no fallback, it only adapts sections).

GauBen:
@delallioui I'll be pushing a commit here, you'll need to rebase

delallioui:
> @delallioui I'll be pushing a commit here, you'll need to rebase

OK (y)

GauBen:
I unified edit and create queries, we'll see if tests still pass. @gflores-jahia opened https://github.com/Jahia/jcontent/issues/2584 to see why the two queries drifted over time

GauBen (approved):

gflores-jahia (approved):

Issue #2514, opened 2026-07-03
Accessing the translate panel through different means produces different interfaces


GauBen:
### Environment and versions used

jContent snapshot (commit 772b7c0fff8fc0d6768b232e4e280d7b2b8cf3fe)

### Steps to reproduce

First way:

Right click > Advanced Edit > Translate

<img width="1296" height="213" alt="Image" src="https://github.com/user-attachments/assets/c115410f-8c10-4158-b735-95f27c1c933d" />

<img width="250" height="148" alt="Image" src="https://github.com/user-attachments/assets/7f9181c1-7862-4935-bbd2-f765754bb463" />

<img width="3228" height="570" alt="Image" src="https://github.com/user-attachments/assets/e418c532-9e7b-4b6b-be52-6728910f5fe6" />

---

Second way:

Right click > Translate

<img width="1293" height="201" alt="Image" src="https://github.com/user-attachments/assets/67e62b9e-e7a7-42ed-a827-2a17075aaca3" />

<img width="1600" height="357" alt="Image" src="https://github.com/user-attachments/assets/8dd7e00e-0e84-4f68-b750-01f3134d8bc6" />

### Current Behavior

When using the first way, some additional categories are added to the editor

### Desired Behavior

Both ways should produce the same interface

PR #2578, opened 2026-07-17
feat: side-by-side diff indicator on the compare column (#2556)

### LINKED ISSUE #2556: Side-by-side compare: highlight fields that differ from the source (diff indicator + gated copy arrow)
### Context

Follow-up from the **content-versioning** version-history UI ([content-versioning#179](https://github.com/Jahia/content-versioning/issues/179), PR [content-versioning#193](https://github.com/Jahia/content-versioning/pull/193)). The version panel reuses jContent's side-by-side machinery (`sideBySideContext` + the per-field `translateField` copy-arrow) to restore a property from a past version into the live edit form — exactly like the Translate tab's source column.

The design ([Figma](https://www.figma.com/design/AENNk51AJJl9748mDTHMNy/Versioning?node-id=4954-15442)) shows the read-only compare column marking **only the fields that differ** from the live value: a **blue vertical bar** on the left edge of each differing field, and the copy/restore arrow shown **only on those differing fields**.

### Problem

jContent currently has **no diff/compare indicator** in the side-by-side UI:

- `FormBuilder/Field/Field.container.jsx` (L25-30) renders `DisplayAction actionKey="translateField"` for every field;
- `actions/contenteditor/translate/translateFieldAction.jsx` (L7-51) renders a single `<ArrowLeft/>` copy button whenever `sideBySideContext.enabled` — it does **not** compare source vs target, so the arrow shows on every valued field, not only differing ones;
- `TranslatePanel/SourceContentPanel.jsx` (L42-50) sets up `sideBySideContext` with no diff flag;
- there is no diff/highlight styling anywhere in the ContentEditor SCSS;
- `utils/fields.utils.js` `checkIfValuesAreDifferent` / `propertyHasChanged` (L213/L230) compare the formik value against saved `nodeData` for **save/dirty detection only** — not against the side-by-side source value, and not wired to any UI.

### Ask

Introduce a **difference indicator** in the side-by-side / translate compare UI:

1. a `sideBySideContext` flag to enable diff mode;
2. per-field detection of "value differs from the compared source" (the `checkIfValuesAreDifferent` helper is the natural primitive, pointed at the source value);
3. the **blue left-bar** visual on differing fields;
4. gate the copy/restore arrow so it appears **only** on differing fields.

Both the Translate tab and content-versioning's version compare would benefit (see also #2514 on translate-panel consistency).

### Notes

- This is a **net-new editor-core feature**, not a versioning-module concern — likely jContent/editor-team owned. Filed here so the version-history "restore only what changed" UX and the Translate tab share one implementation.
- Until this lands, content-versioning ships restore on every valued i18n field (documented limitation on PR #193).

Area: Tech · javascript

delallioui:
Closes #2556

### What
Add an opt-in **difference indicator** to the side-by-side compare UI, gated behind a new **`sideBySideContext.showDiff`** flag:
- a **blue left-bar** on each compare-column field whose value differs from the compared live node;
- the restore/copy **arrow shown only on differing fields** (in diff mode).

### Why
The compare UI (Translate tab, and content-versioning's version panel which reuses `sideBySideContext` + the per-field `translateField` arrow) had **no diff mechanism**: the arrow showed on every valued i18n field, with no cue for what actually changed. The design ([Figma](https://www.figma.com/design/AENNk51AJJl9748mDTHMNy/Versioning?node-id=4954-15442)) marks only differing fields with a blue bar + restore arrow.

### How
- `FieldContainer` computes `hasDiff` via the existing `propertyHasChanged(sourceValue, field, sideBySideContext.targetNodeData)` primitive (compares the compare-column value against the live node's **saved** value), applies a `.fieldContainerDiff` bar class, and passes `hasDiff` to the `translateField` action.
- `translateFieldAction` renders the arrow only when `!showDiff || hasDiff`.
- New `.fieldContainerDiff` SCSS (blue `var(--color-accent)` left-bar).

### Behaviour / scope
**Fully gated — no change anywhere today.** With `showDiff` unset (the main edit form and the current Translate tab) behaviour is identical: no bar, arrow on every valued field. Consumers opt in by setting `showDiff` + `targetNodeData`:
- **content-versioning's version panel** (separate change) — resolves its documented "restore on every valued field" limitation on PR #193;
- the **Translate tab** could opt in under #2514 (deliberately not flipped here, to avoid changing shipped Translate behaviour).

Design decision: compare against the live node's **saved** value (reachable, reuses `propertyHasChanged`); it doesn't reflect unsaved live edits until save — acceptable for the version-compare "what changed since this version" case.

### Verification
- New `translateFieldAction.spec.jsx` — 9 tests (arrow gating: disabled/non-i18n/no-lang → none; translate mode unchanged incl. hasDiff=false; diff mode → arrow only when differs; click still copies).
- New `Field.container.diff.spec.jsx` — 3 tests (bar class + `hasDiff` prop; inert when `showDiff` off).
- Existing `Field.container.spec.jsx` unchanged (4).
- Full ContentEditor jest suite green (578 tests); ESLint + Stylelint clean.

---
Closes Jahia/content-versioning#179

### DIFF
--- a/src/javascript/ContentEditor/actions/contenteditor/translate/translateFieldAction.jsx
+++ b/src/javascript/ContentEditor/actions/contenteditor/translate/translateFieldAction.jsx
@@ -4,12 +4,19 @@ import {useContentEditorConfigContext, useContentEditorContext} from '~/ContentE
 import {ArrowLeft} from '@jahia/moonstone';
 import styles from './styles.scss';
 
-export const TranslateFieldActionComponent = ({field, value, render: Render}) => {
+export const TranslateFieldActionComponent = ({field, value, hasDiff, render: Render}) => {
     const {sideBySideContext} = useContentEditorConfigContext();
     const {setI18nContext} = useContentEditorContext();
 
-    const {enabled, translateLang, hasWritePermission, lockedAndCannotBeEdited} = sideBySideContext || {};
+    const {enabled, translateLang, hasWritePermission, lockedAndCannotBeEdited, showDiff} = sideBySideContext || {};
     if (!enabled || !field.i18n || !translateLang) {
+        return;
+    }
+
+    // In diff mode (issue #2556) the restore arrow is shown only where the value differs from the
+    // compared node — copying an identical value would be a no-op. Outside diff mode the arrow keeps
+    // showing on every valued field (unchanged Translate-tab behaviour).
+    if (showDiff && !hasDiff) {
         return;
     }
 
@@ -53,6 +60,7 @@ export const TranslateFieldActionComponent = ({field, value, render: Render}) =>
 TranslateFieldActionComponent.propTypes = {
     field: PropTypes.object.isRequired,
     value: PropTypes.any,
+    hasDiff: PropTypes.bool,
     render: PropTypes.func.isRequired
 };
 
--- a/src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/styles.scss
+++ b/src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/styles.scss
@@ -6,3 +6,22 @@
     flex-direction: row;
     align-items: stretch;
 }
+
+// Side-by-side diff mode (issue #2556): a blue vertical bar on the left edge of a field whose value
+// differs from the compared node, marking what changed in the read-only compare column.
+.fieldContainerDiff {
+    padding-left: var(--spacing-small);
+
+    &::before {
+        content: '';
+
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        left: 0;
+
+        width: 2px;
+
+        background-color: var(--color-accent);
+    }
+}
--- a/src/javascript/ContentEditor/actions/contenteditor/translate/translateFieldAction.spec.jsx
+++ b/src/javascript/ContentEditor/actions/contenteditor/translate/translateFieldAction.spec.jsx
@@ -1,0 +1,86 @@
+import React from 'react';
+import {shallow} from '@jahia/test-framework';
+import {TranslateFieldActionComponent} from './translateFieldAction';
+import {useContentEditorConfigContext, useContentEditorContext} from '~/ContentEditor/contexts';
+
+jest.mock('~/ContentEditor/contexts', () => ({
+    useContentEditorConfigContext: jest.fn(),
+    useContentEditorContext: jest.fn()
+}));
+
+// A no-op stand-in for the button renderer; we only assert whether/how the arrow is rendered.
+const Render = () => null;
+
+describe('TranslateFieldActionComponent', () => {
+    const i18nField = {name: 'nt_prop', i18n: true};
+    let setI18nContext;
+
+    const render = (sideBySideContext, props = {}) => {
+        useContentEditorConfigContext.mockReturnValue({sideBySideContext});
+        useContentEditorContext.mockReturnValue({setI18nContext});
+        return shallow(
+            <TranslateFieldActionComponent field={i18nField} value="source" render={Render} {...props}/>
+        );
+    };
+
+    // Whether the restore arrow is rendered at all.
+    const hasArrow = wrapper => wrapper.find(Render).exists();
+
+    beforeEach(() => {
+        setI18nContext = jest.fn();
+    });
+
+    it('renders no arrow when side-by-side is disabled', () => {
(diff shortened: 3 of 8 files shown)


src/javascript/ContentEditor/actions/contenteditor/translate/translateFieldAction.jsx
```diff
@@ -4,15 +4,22 @@ import {useContentEditorConfigContext, useContentEditorContext} from '~/ContentE
...
-    const {enabled, translateLang, hasWritePermission, lockedAndCannotBeEdited} = sideBySideContext || {};
+    const {enabled, translateLang, hasWritePermission, lockedAndCannotBeEdited, showDiff} = sideBySideContext || {};
     if (!enabled || !field.i18n || !translateLang) {
         return;
     }
 
+    // In diff mode (issue #2556) the restore arrow is shown only where the value differs from the
+    // compared node — copying an identical value would be a no-op. Outside diff mode the arrow keeps
+    // showing on every valued field (unchanged Translate-tab behaviour).
+    if (showDiff && !hasDiff) {
```
eallais, line 19: When the compare-column value is empty but the live node has one, `propertyHasChanged` reports a diff (correctly — it did change), so the bar shows and this gate lets the arrow through. But the arrow's `enabled` is `Boolean(value) && …`, so it renders disabled: the field is flagged as "differs" with no way to act on it via the arrow.<br><br>• In the version-restore case (a property added after the compared version), "restore" would mean clearing the field — which the copy arrow can't express (its onClick is a no-op when `value` is empty).<br>• Not a blocker — just want to confirm the empty→value direction is intentionally out of scope rather than an overlooked case.<br><br>_Source: the `enabled` prop a few lines below — `enabled={Boolean(value) && hasWritePermission && !lockedAndCannotBeEdited}`._

src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/styles.scss
```diff
@@ -6,3 +6,22 @@
...
+        content: '';
+
+        position: absolute;
+        top: 0;
+        bottom: 0;
+        left: 0;
+
+        width: 2px;
+
+        background-color: var(--color-accent);
```
eallais, line 25: The diff bar communicates "this field changed" through color alone (blue `--color-accent`). Where the restore arrow isn't a redundant cue — non-i18n fields have no arrow at all — a color-blind or low-vision user has nothing to tell a changed field from an unchanged one.<br><br>• WCAG 1.4.1 (Use of Color) asks for a second, non-color cue for anything conveyed by color.<br>• A patterned/thicker bar, a small icon, or an `aria`/`title` hint on the flagged field would cover it cheaply.<br><br>Is a non-color affordance in scope, or is the bar-only treatment fixed by the Figma design?<br><br>_Source: [WCAG 2.1 — 1.4.1 Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html)._

gflores-jahia (dismissed):

src/javascript/ContentEditor/actions/contenteditor/translate/translateFieldAction.jsx
delallioui: Confirmed — intentional and out of scope. The copy arrow only *copies a value into* the live form; its `onClick` is a no-op when the compare-column value is empty (`return (value) ? result : prev`), so it structurally can't express "clear the field". Restoring an empty compare value (a property added after the compared version) would mean clearing, which this mechanism doesn't do.

The bar is an honest "this field differs" signal, so we keep it shown while the arrow stays disabled rather than hiding a real difference. The empty→value / clear direction is deliberately left to a future change if it's ever needed.

src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/styles.scss
delallioui: Good catch — added in 75211b8e. Flagged fields now carry a localised `title`/tooltip ("Value differs from the compared version", en/fr/de) surfaced to assistive tech, so the difference is perceivable without relying on the bar's colour. This specifically covers non-i18n fields, where the restore arrow (the redundant cue on i18n fields) never renders.

The bar's visual treatment itself stays as the Figma design specifies; the title is the non-colour second cue on top of it.

GauBen:
We cannot build on top of the translation actions, we need a proper abstraction

Closing until we figure out a plan

PR #2580, opened 2026-07-17
fix: replace selected domain from resolved live url

### LINKED ISSUE #2115: Live button with several server names
**As a content author**
**I want to go to my live website using different domains**
**In order to test my website with or without CDN**

### Context / Problem to solve
When a CDN is active, I want to have a link to my website without the CDN to test my changes

### Ideas / Already identified solutions 
if several server names are configured, update the live button to allow users to select different server names. 
Users must not need to select the server name each time, but rather select once and they we keep that information in the local storage. 

Challenge: How do we tell users:
- which server name is the default one 
- which server name is currently selected in the local storage  / which server name is currently used by the link button

gflores-jahia:
### Description

Fixed `resolveUrlForLiveOrPreview` in `JContent.utils.jsx` to correctly handle localhost selection, alias substitution, and port-only-on-local-domain. 

After fix:

| Selected server | Result |
|---|---|
| Current domain (`serverName === location.hostname`) | `location.hostname:PORT` — port preserved |
| `localhost` | `localhost:PORT` — port preserved |
| External server name (alias or primary, non-localhost) | Replace returned absolute url with selected`serverName` — no port, standard 80/443 assumed |

Jest tests added and passing. 

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/JContent/JContent.utils.jsx
+++ b/src/javascript/JContent/JContent.utils.jsx
@@ -467,20 +467,27 @@ export const JahiaRenderedModulesUtil = {
 };
 
 export const resolveUrlForLiveOrPreview = (url, isLive, serverName) => {
-    if (!url || url.match(/^https*:\/\//)) {
+    if (!url) {
         return url;
     }
 
-    // Use current host for preview urls
-    let host = location.hostname;
-    let port = location.port ? `:${location.port}` : '';
+    // Strip host from absolute URL — path is identical across all server names for the same site
+    const path = url.replace(/^https*:\/\/[^/]+/, '');
+    const port = location.port ? `:${location.port}` : '';
 
-    if (isLive && serverName && serverName !== 'localhost' && serverName !== location.hostname) {
-        host = serverName;
-        port = '';
+    const isCurrentDomain = !isLive || !serverName || serverName === location.hostname;
+    const isLocalServer = serverName === 'localhost';
+
+    if (isCurrentDomain) {
+        return `${location.protocol}//${location.hostname}${port}${path}`;
     }
 
-    return `${location.protocol}//${host}${port}${url}`;
+    if (isLocalServer) {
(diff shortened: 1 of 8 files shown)


src/javascript/JContent/JContent.utils.test.jsx
```diff
@@ -36,6 +36,165 @@ describe('getNewCounter', () => {
...
+
+    const relativePath = '/cms/render/live/en/sites/digitall/home.html';
+    const absoluteServera = 'http://servera/cms/render/live/en/sites/luxe/home.html';
+    const absoluteServeraWithPort = 'http://servera:8080/cms/render/live/en/sites/luxe/home.html';
+    const absoluteServerb = 'http://serverb/cms/render/live/en/sites/siteb/home.html';
+
+    const setLocation = (hostname, port = '') => {
+        delete globalThis.location;
+        globalThis.location = {hostname, port, protocol: 'http:'};
+    };
```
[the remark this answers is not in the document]
gflores-jahia: Updated and resolved

src/javascript/JContent/JContent.utils.jsx
```diff
@@ -467,20 +467,28 @@ export const JahiaRenderedModulesUtil = {
 };
 
 export const resolveUrlForLiveOrPreview = (url, isLive, serverName) => {
-    if (!url || url.match(/^https*:\/\//)) {
+    if (!url) {
```
AKarmanov: I find this a little odd. Should the user even be calling this function if he knows he has no url, is there a use case for this scenario?

AKarmanov (approved):

src/javascript/JContent/JContent.utils.jsx
gflores-jahia: I think you're right, url is always constructed for the path in theory. I'm moving the check to Open live button instead to handle any other exception cases just in case and not completely crash jcontent.

PR #2581, opened 2026-07-20
feat: supported side-by-side source seam (SideBySideSourceProvider) (#2579)

### LINKED ISSUE #2579: Provide a supported seam for restoring a property from a read-only side-by-side source
## Context

The content-versioning module renders a side-by-side version panel where a read-only "version" column sits next to the live edit form. To let a user **restore a property** from a version, we need jContent's per-field copy-arrow (the one the Translate tab shows) to appear on the read-only column.

Today the only way a consuming module can light up that arrow is to fake a "side-by-side source" by injecting a `sideBySideContext` object into the ContentEditor config:

```tsx
// content-versioning: src/javascript/components/VersionSnapshotForm/index.tsx
const restoreConfig = useMemo(() => ({
    ...editorConfig,
    sideBySideContext: {
        ...editorConfig.sideBySideContext,
        enabled: true,
        readOnly: true,
        translateLang: lang,
        hasWritePermission: editorContext.nodeData?.hasWritePermission,
        lockedAndCannotBeEdited: editorContext.nodeData?.lockedAndCannotBeEdited
    }
}), [...]);
```

This works, but it's a workaround: the module is reaching into jContent's internal translate/side-by-side machinery, the field slot in `FieldContainer` is hardcoded to `translateField`, and the arrow only decorates i18n fields (so shared / non-i18n properties can't be restored this way).

## Ask

Provide a supported jContent seam for "restore a value into the edit form from an arbitrary read-only source column" — i.e. a first-class side-by-side source configuration (or a dedicated restore action) that doesn't require modules to hand-craft `sideBySideContext`, and that could eventually cover shared (non-i18n) properties too.

## References

- Raised in content-versioning PR review: https://github.com/Jahia/content-versioning/pull/193 (comment by @GauBen: "this should be addressed in jcontent, we won't live with weird workarounds forever")
- Consuming code: `content-versioning` → `src/javascript/components/VersionSnapshotForm/index.tsx`
- Related product issue: Jahia/content-versioning#179

delallioui:
## What

Adds `SideBySideSourceProvider` — a supported jContent seam for rendering a **read-only side-by-side source column** next to the live edit form, so consuming modules no longer have to hand-craft the internal `sideBySideContext` object.

Motivated by content-versioning's version panel, which restores a value from a past version by faking a side-by-side source (see the [PR review](https://github.com/Jahia/content-versioning/pull/193) and Jahia/content-versioning#179). jContent's own Translate tab was building the *same* object independently.

## Changes

- **New `SideBySideSourceProvider`** (`contexts/ContentEditorConfig/`) — reads the surrounding config + node contexts, builds and memoizes `sideBySideContext` (`enabled`/`readOnly`/`translateLang` + permissions taken from the editable node), and provides the overridden config down the tree. Named `translateLang` / `sourceLang` props; the defaults reproduce the Translate tab's behaviour. Exported through the federation barrel (`@jahia/jcontent`).
- **Dogfood** — `SourceContentPanel` (Translate tab) now uses the provider instead of hand-building the context; behaviour-preserving (89 → 16 lines).

Consumers go from hand-crafting the context to:

```tsx
<SideBySideSourceProvider translateLang={lang}>
    <FormBuilder mode="edit"/>
</SideBySideSourceProvider>
```

## Out of scope (intentionally)

- The issue's *eventual* goals — restoring **shared (non-i18n)** properties, and a dedicated restore action — are not built here. The seam is forward-compatible with them, and with #2556's `showDiff`/`targetNodeData` (add named props once that lands).
- The **content-versioning consumer wiring** is a separate follow-up, blocked on a released jContent version bumped in its deps.

## Tests

- New `SideBySideSourceProvider.spec.jsx` (built config + prop overrides + base-config preservation) and `SourceContentPanel.spec.jsx` (dogfood wiring). Affected + surrounding suites green; eslint clean; webpack build compiles.

Closes #2579

### DIFF
--- a/src/javascript/ContentEditor/contexts/ContentEditorConfig/SideBySideSourceProvider.jsx
+++ b/src/javascript/ContentEditor/contexts/ContentEditorConfig/SideBySideSourceProvider.jsx
@@ -1,0 +1,51 @@
+import React, {useMemo} from 'react';
+import * as PropTypes from 'prop-types';
+import {ContentEditorConfigContextProvider, useContentEditorConfigContext} from './ContentEditorConfig.context';
+import {useContentEditorContext} from '~/ContentEditor/contexts/ContentEditor';
+
+/**
+ * Renders its children as a read-only "side-by-side source" column, next to the live edit form — the
+ * setup the Translate tab uses to show a source language, and the supported seam for modules that want
+ * the same thing (e.g. content-versioning restoring a value from a past version). See issue #2579.
+ *
+ * It reads the surrounding config and node contexts, builds the `sideBySideContext` (so consumers no
+ * longer hand-craft it), and provides an overridden config down the tree. Per-field restore arrows
+ * (`translateFieldAction`) light up for i18n fields, copying the source value into the live form's
+ * `translateLang`.
+ *
+ * @param translateLang the editable language a restored value is copied into; defaults to the current config language
+ * @param sourceLang    the language this source column renders in; defaults to the existing side-by-side language, then the config language
+ */
+export const SideBySideSourceProvider = ({translateLang, sourceLang, children}) => {
+    const baseConfig = useContentEditorConfigContext();
+    const {nodeData} = useContentEditorContext();
+
+    const hasWritePermission = nodeData?.hasWritePermission;
+    const lockedAndCannotBeEdited = nodeData?.lockedAndCannotBeEdited;
+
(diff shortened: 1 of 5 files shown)


GauBen:
> The issue's eventual goals — restoring shared (non-i18n) properties, and a dedicated restore action — are not built here

Closing to reduce noise, as this PR does not achieve its goal

PR #2590, opened 2026-07-21
test: move visibility test from selenium to cypress

sdusolle:
### Description
Add scenarios on visibility conditions:

- copy/paste a richtext with a condition=visible
- copy/paste a richtext with a condition= non visible
- add a content reference to a richtext with a condition=visible
- add a content reference to a richtext with a condition= non visible

This replace Selenium test `visibilityConditionsOnCopiedReferenceTest`.

#### Source code

#### Tests

> [!TIP]


danielaebenberger (commented):
Thanks for migrating the tests. Some feedback found by the /pr-test-reviewer skill from qa-tool:

**Bug:** "referenced content with visible condition is shown in live" (test 3) is a false-positive risk. The test right before it (test 2) pastes a copy of the same visible richtext into `area-main` and publishes — nothing cleans that up between `it()`s (cleanup only runs once, in the describe's `after()`). So by test 3, that text is already rendering on the page, independent of whether the new reference node's visibility condition evaluates correctly at all. Even a fully broken reference-visibility feature would pass this test as written.
→ Fix: delete/unpublish prior tests' pasted nodes first, or put the reference scenarios on their own page/area. (The negative test right after isn't affected.)

**Other issue:**
- `createVisibilityCondition.graphql` duplicates the existing (unused) `createRules.graphql` in the same folder — same `jnt:conditionalVisibility`/`dayOfWeekCondition` structure. Parameterize and reuse that one instead of adding a sibling.
- PR description says "add a content reference" via copy/paste, but the implementation creates the reference node directly via GraphQL — the actual "Paste as reference" UI button (used elsewhere in `pageBuilder/clipboard.cy.ts`) is never exercised. Is this narrower scope intended?

sdusolle:
Thanks for the review!

**1. False-positive risk in test 3** (referenced content with visible condition is shown in live) — fixed by moving the two reference scenarios onto their own dedicated page:

- Added refPageName = 'testPageVisibilityRef' and refAreaMainPath constants.
- before() now creates a second jnt:page with its own area-main; after() deletes it.
- Both reference tests (ref-visible and ref-hidden) now add their jnt:contentReference to refAreaMainPath, publish refPageName, and assert against refPageName.html.

Because the copy/paste tests only ever write into testPageName's area, nothing they leave behind renders on the reference page. Test 3 now genuinely fails if reference-visibility evaluation is broken. (The negative test remains correct — it only asserts the hidden text is absent, and the visible reference sharing the page doesn't interfere.)

**2. Duplicate GraphQL fixture** — consolidated:

- Rewrote `createRules.graphql` as a parameterized mutation createRules($contentPath: String!, $dayOfWeek: [String]!).
- Deleted the sibling createVisibilityCondition.graphql.
- Updated both cy.apollo calls in the test to reference createRules.graphql.

**3. PR description** says "add a content reference" via copy/paste, but the implementation creates the reference node directly via GraphQL --> the goal of this testcase here is not to test the creation of the reference but only that the visibility condition is well applied. So I would keep it that way.

gflores-jahia (approved):

danielaebenberger (approved):

PR #2591, opened 2026-07-21
fix: refresh handlers after switching CE language

r3dm1ke:
### Description
This PR fixes a bug where the language is switched in CE form, and some pickers loose internal state.

#### Source code

#### Tests

> [!TIP]

### DIFF
--- a/src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/Field.jsx
+++ b/src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/Field.jsx
@@ -250,6 +250,22 @@ export const Field = ({inputContext, idInput, selectorType, field}) => {
 
     const currentValue = values[field.name];
 
+    // Sections are replaced with a fresh server copy when the form reloads without remounting
+    // (language switch, see ContentEditorSection.context). That copy is computed from persisted
+    // state only, so it loses client-side handler effects (dependent valueConstraints, addMixin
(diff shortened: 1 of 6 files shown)


src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/Field.jsx
```diff
@@ -250,6 +250,22 @@ export const Field = ({inputContext, idInput, selectorType, field}) => {
...
+    const lastSectionsRef = useRef(sectionsContext.sections);
+    useEffect(() => {
+        if (lastSectionsRef.current !== sectionsContext.sections) {
+            lastSectionsRef.current = sectionsContext.sections;
+            onChangeValue.current = undefined;
+            if (currentValue !== null && currentValue !== undefined) {
+                registeredOnChangeRef.current(currentValue);
+            }
+        }
+    }, [sectionsContext.sections, currentValue]);
```
gflores-jahia: This looks ok for now, but there is a potential for this somewhere down the road to be switched out of order with the other useEffect that also fires currentValue and can alter the final value read and associated previousValue:

```
useEffect(() => {
    registeredOnChangeRef.current(currentValue);
}, [currentValue]);
```

I'd prefer if we combine these two to guarantee the ordering and also simplify the useEffects.

We should also move this below the count useEffect below as it could also lead to stale data refresh and that the currentValue useEffect would always be more updated than initialValue mount.

gflores-jahia (changes requested):

sdusolle (dismissed):

PR #2613, opened 2026-07-23
fix: improve the mechanism responsible for picking the source language in translation layout

### LINKED ISSUE #2484: Source and target language values for translate mode

- Latest snap
- Ensure your site has EN, SP and FR as active languages. EN being the default language. 

## Scenario 1: When switching from edit to translate 

- Advanced edit a content in EN
- Change mode to translate 
- The target language should be EN (already the case)
- The source language should be calculated as follow (not sure about which condition should be executed first but here is the logic):
-- Take all languages that are active in edit on the website and that have an existing translation 
-- Remove the target language (there is not point to translate to the same language)
-- If possible, select the default language
-- If not, select the first language in alphabetical order
-- If no language matches, fallback to the current lang

=> If I'm editing in EN, and if there are translations existing for french and spanish => select french because alphabetical
=> If I'm editing in FR, and if there is a translation existing for english and spanish => select english because it's default 
... 

## Scenario 2: When clicking translate in the right click actions 
When right clicking on a content item in page builder, clicking "Translate" (**change label to "Translate to"**) has a somehow different behavior, it should be understood as "Translate from the current language to others" => 

If the site has EN, SP and FR as active languages. EN being the default language. 
Given I'm in EN in page builder
When using the right click in page builder on a content
Then I go to the translate mode 
With source language as english (the source language is easy, it's the same as the language in PB)
And target language is calculated as the following: first language without a translation that is different from target language, in alphabetical order.

GauBen:
Closes #2484

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@apollo/react-components": "^4.0.0",
     "@apollo/react-hoc": "^4.0.0",
     "@babel/polyfill": "^7.11.5",
-    "@jahia/data-helper": "^1.1.19",
+    "@jahia/data-helper": "^1.1.21",
     "@jahia/design-system-kit": "^1.2.1",
     "@jahia/icons": "^1.1.2",
     "@jahia/moonstone": "^2.20.3",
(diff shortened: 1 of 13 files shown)


sdusolle (dismissed):

AKarmanov (dismissed):

AKarmanov:
I'm not sure about the current jcontent release scope, we were discussing this with @rknj yesterday, so maybe just confirm that this is part of the up coming release before merging.

rknj:
Yes we're OK to push this in 3.7.0, we've discussed this yesterday with @romain-pm and @GauBen

GauBen:
Was previously approved and only made minimal changes to fix failing/flaky tests, bypassing a second round of reviews

PR #2375, opened 2026-06-05
test: Fix failing test

AKarmanov:
### Description
Fix failing test.

#### Source code
- [ ] I've shared and documented any breaking change
- [ ] I've reviewed and updated the jahia-depends

#### Tests
- [ ] I've provided Unit and/or Integration Tests
- [ ] I've updated the parent issue with required manual validations

> [!TIP]
> Documentation to guide the reviews: [How to do a code review](https://jahia-confluence.atlassian.net/wiki/spaces/PR/pages/2064660/How+to+do+a+code+review+-+Ref+ISSOP08.A14006)

PR #2622, opened 2026-07-27
fix(preview): render previewed content as a static document (3_6_x)

delallioui:
Maintenance-line counterpart of #2621 (`main`), applied to the `3_6_x` layout.

## Summary

The preview panels render a node's `renderedContent` output into an iframe and let that markup run
scripts of its own inside the application's origin. The preview is meant to show **layout**: only
stylesheets are added to the frame and the frame body is made non-interactive once the content is in,
so scripts belonging to the previewed content were never part of the intended rendering.

On this line the preview exists in **two** places, and they had drifted apart:

| Frame | How content gets in | Sandbox before |
|---|---|---|
| Content Editor preview (`PreviewViewers/IframeViewer.jsx`) | `srcDoc` | `allow-same-origin allow-scripts` |
| jContent preview drawer (`PreviewComponent.jsx`) | `document.write` | **none declared at all** |

The second frame declared no `sandbox` attribute, so it was not merely script-enabled — it was
entirely unrestricted (forms, popups and top-level navigation included).

## Change

- Both preview frames now declare `sandbox="allow-same-origin"`, so each presents the rendered markup
  as a static document. Same value, same named `PREVIEW_SANDBOX` constant, in both files.
- `allow-same-origin` is deliberately kept in both. The component writes the content into the frame
  and adds its stylesheets by reaching into the frame's `document`, which only works while the frame
  keeps the application's origin. Dropping it would silently cost the preview its content and CSS.
- In `PreviewComponent`, `sandbox` is applied **after** the `{...iframeProps}` spread so a caller
  cannot widen it by accident. No current caller passes one.

### Behaviour change

A previewed component's own scripts no longer run inside either preview frame. Layout and CSS are
unaffected, but a template that relies on JavaScript to finish its presentation (a carousel,
JS-computed sizing, lazy-loaded media) will show its pre-script state in the preview.

Only the embedded preview frames are affected — Page Composer, the live site and the standalone
preview window are unchanged. No configuration or API changes; nothing to migrate.

## Verification

Unit tests on this line, each failing before the change and passing after:

- `PreviewViewers/IframeViewer/IframViewer.spec.js` — the Content Editor frame is declared without
  `allow-scripts` and with `allow-same-origin`.
- `PreviewComponent/PreviewComponent.spec.js` (new) — the jContent frame carries the same value
  (previously `undefined`), and a caller-supplied `iframeProps.sandbox` cannot widen it.

`yarn jest` over both preview trees: 23 tests passing, lint clean. The end-to-end coverage for this
behaviour lives on `main` (#2621), where the two frames are already unified into one component.

### DIFF
--- a/.chachalog/KnNfP.md
+++ b/.chachalog/KnNfP.md
@@ -1,0 +1,5 @@
+---
+jcontent: patch
+---
+
+Preview renders content as a static document: scripts embedded in previewed content no longer run in the preview frames
--- a/.chachalog/KnNfP.md
+++ b/.chachalog/KnNfP.md
@@ -1,5 +1,6 @@
 ---
-jcontent: patch
+# Allowed version bumps: patch, minor, major
+"@jahia/jcontent": patch
 ---
 
 Preview renders content as a static document: scripts embedded in previewed content no longer run in the preview frames
--- a/.chachalog/KnNfP.md
+++ b/.chachalog/KnNfP.md
@@ -3,4 +3,4 @@
 "@jahia/jcontent": patch
 ---
 
-Preview renders content as a static document: scripts embedded in previewed content no longer run in the preview frames
+Render previewed content as a static document: embedded scripts no longer run in the preview frames (#2622)
--- a/src/javascript/ContentEditor/editorTabs/EditPanelContent/Preview/PreviewViewers/IframeViewer/IframViewer.spec.js
+++ b/src/javascript/ContentEditor/editorTabs/EditPanelContent/Preview/PreviewViewers/IframeViewer/IframViewer.spec.js
@@ -28,6 +28,17 @@ describe('IframeViewer', () => {
         const cmp = shallow(<IframeViewer {...props}/>);
         expect(cmp.find('LoaderOverlay').exists()).toBe(true);
     });
+
+    it('should render the rendered content as a static document, without running its scripts', () => {
+        const cmp = shallow(<IframeViewer {...props}/>);
+        const sandbox = cmp.find('iframe').prop('sandbox');
+
+        // The preview shows layout only - the framed document must not execute scripts.
+        expect(sandbox).not.toContain('allow-scripts');
+        // ...but it must stay same-origin, otherwise loadAssets and zoom can no longer reach
+        // the frame's document and the preview loses its CSS and its zoom behaviour.
+        expect(sandbox).toContain('allow-same-origin');
+    });
 });
 
 describe('Zoom test', () => {
--- a/src/javascript/ContentEditor/editorTabs/EditPanelContent/Preview/PreviewViewers/IframeViewer/IframeViewer.jsx
+++ b/src/javascript/ContentEditor/editorTabs/EditPanelContent/Preview/PreviewViewers/IframeViewer/IframeViewer.jsx
@@ -23,6 +23,18 @@ export function zoom(iframeDocument, onContentNotFound, editorContext) {
     }
 }
 
+/**
+ * Sandbox flags for the preview frame.
+ *
+ * The preview is a *static layout* rendering of a node: only stylesheets are added to the frame and
+ * its body is made non-interactive on load (`pointer-events: none`), so the framed document never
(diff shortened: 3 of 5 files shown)


delallioui:
*Superseded — see the consolidated summary below.* (This comment blamed the moving `8-SNAPSHOT` image; that turned out to be wrong.)

delallioui:
*Superseded — see the consolidated summary below.*

GauBen (approved):

delallioui:
### Why the checks are red (consolidated)

**Not caused by this change, and not by the Jahia image.** I ran the four preview specs behind the failing suites in four environments:

| Environment | `8-SNAPSHOT` | `8.2.3-SNAPSHOT` |
|---|---|---|
| without the Digitall import | 15 fail | 15 fail |
| with the full provisioning | **21 pass** | **21 pass** |

Platform version is irrelevant; the failures track the *test environment*. (My earlier "drift from the moving image" comments were wrong — apologies for the noise.)

**Actual cause:** CI logs are full of federation mismatches — jcontent 3.6.2 ships react 18.3.1 / react-i18next 11.18.6 / data-helper 1.1.21, while the companion modules `provisioning.yml` installs (`legacy-default-components`, `dx-base-demo-*`, `qa-module`) require `^16.x` / `^13-14` / `=1.0.12`. Those are installed **unpinned**, so they drifted forward since the last green PR onto this branch ([#2375](https://github.com/Jahia/jcontent/pull/2375), early June). Fixing that means pinning them for this line — a maintainer call, out of scope here.

Scale, for the record: on the `jcontent` profile it's **4 specs of 26** (`preview`, `links`, `selection`, `compareStagingLive`), not a red suite.

**SonarQube** is a separate, unrelated misconfiguration: this branch passes `primary_release_branch: master`, and no `master` branch exists — the action feeds it to `-Dsonar.pullrequest.base`, which defines *new code*, so the whole branch got attributed to this PR (19 reliability + 83 maintainability issues, on files never touched). Fix is one line, tracked separately. The only genuinely attributable findings were two `[S1090] iFrames must have a title`, now fixed on both frames to match `main`.

**What is verified here:** unit assertions red before / green after on both frames (23 tests); the preview specs pass in a faithful environment; the built bundle contains zero `allow-scripts`.

gflores-jahia:
there are failing tests in menu-actions and jcontent regarding preview that might be related to the change and worth looking at. 

At least an assessment of reason why it's failing and if it can be reproduced locally can be a start, we can check if related to the change or not. 

https://github.com/Jahia/jcontent/runs/90031408520#r11

```
test preview in content tab
 |   |    | - FAIL: Menu actions preview tests test preview in content tab shows preview on content folder item
 |   | - test preview in media tab
 |   |    | - FAIL: Menu actions preview tests test preview in media tab shows preview on PDF
 |   |    | - FAIL: Menu actions preview tests test preview in media tab shows preview on files
 |   |    | - FAIL: Menu actions preview tests test preview in media tab can select another item
 |   |    | - FAIL: Menu actions preview tests test preview in media tab can switch to list view
 |   |    | - FAIL: Menu actions preview tests test preview in media tab can select another item
 |   | - Test preview in pages tab
 |   |    | - FAIL: Menu actions preview tests Test preview in pages tab Can close preview when using pagination
```

https://github.com/Jahia/jcontent/runs/90032577200#r16

```
 | Suite: Mocha Tests - Total tests: 3 - Failure: 3 - Executed in 89s
 |   | - JContent preview tests
 |   |    | - FAIL: JContent preview tests should honor the j:view property when previewing content
 |   |    | - FAIL: JContent preview tests should show cm view if available and no default
 |   |    | - FAIL: JContent preview tests should show correct preview version in live and edit
```

gflores-jahia:
Failing tests in this comment https://github.com/Jahia/jcontent/pull/2622#issuecomment-5183995242 now fixed with latest test cleanup in 3_6_x branch.

Note about (expected) existing failing tests after the test cleanup yesterday https://github.com/Jahia/jcontent/pull/2708#issuecomment-5343082631.

Should be good to merge.

PR #2634, opened 2026-08-03
feat: expose adaptSections and side-by-side action target

GauBen:
- new `content-editor/field/leading-actions` action target for this gizmo

  image

- exposed `adaptSections` to the federation (not the best thing we can do, but cutting corners here)


src/javascript/ContentEditor/contexts/ContentEditor/ContentEditor.context.tsx
```diff
@@ -13,7 +13,10 @@ import {useOnBeforeContextHooks} from '~/ContentEditor/ContentEditor/useOnBefore
 import {isEqual} from 'lodash';
 import {cloneDeep} from 'es-toolkit';
 
-export const ContentEditorContext = React.createContext({});
+export const ContentEditorContext = React.createContext<{
```
gflores-jahia, line 16: Partial implem of types being passed to context even though there are many more usages. Can this be potentally misleading for IDE users ? I don't know if there's a way to mark rest as any or just keep this ambiguous unless we want to add all usage. I don't know what the best solution here.

src/javascript/ContentEditor/utils/fields.utils.ts
```diff
@@ -290,11 +292,11 @@ export function getDynamicFieldSetNameOfField(sections, sourceField) {
...
  *
- * @param {object} node node to check if it has mixin or not
- * @param {string} mixin mixin name
- * @returns {boolean} true if the node has mixin, false otherwise.
+ * @param node node to check if it has mixin or not
+ * @param mixin mixin name
+ * @returns true if the node has mixin, false otherwise.
  */
-function hasNodeMixin(node, mixin) {
+function hasNodeMixin(node: any, mixin: string): boolean {
```
gflores-jahia, line 298: [nitpick] boolean return type added but can return undefined. Maybe wrap result as Boolean to avoid undefined return. Also simplify with optional chain preferrably.

src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/Field.container.tsx
```diff
@@ -22,12 +22,18 @@ export const FieldContainer = React.memo(({field, inputContext}) => {
...
-            {context.displayActions && <DisplayAction
-                actionKey="translateField"
-                render={ButtonRendererNoLabel}
-                field={field}
-                value={values[field.name]}
-            />}
+            {context.displayActions && (
+                <div className={styles.leadingActions}>
+                    <DisplayActions
+                        target="content-editor/field/leading-actions"
```
gflores-jahia, line 28: Not sure if needed to address yet, but could be a potential perf hit for scanning the registry for each field. If it becomes a perf bottleneck, we can maybe prefetch these actions once, then map and revert back to <DisplayAction>

src/javascript/ContentEditor/contexts/ContentEditor/ContentEditor.context.tsx
GauBen: I think our best bet would be:

- Incremental additions (add the type when you need it) / for 6 months
- Then split into specialized contexts (e.g. `sideBySideContext` deserves its own `useSideBySideContext`) for better safety / afterwards
- Make `use...Context()` throw when not in context, instead of returning an empty object / long-term goal

This way we get full type-safety + a typed API for consumers

src/javascript/ContentEditor/utils/fields.utils.ts
GauBen: I need to address all sonar issues now that I've touched the file, I'll do that as well

src/javascript/ContentEditor/editorTabs/EditPanelContent/FormBuilder/Field/Field.container.tsx
GauBen: No hit on a mac m4, but that's not the baseline 😄

To be monitored closely, I don't think that'll be an issue but I might be proven wrong on older laptops

gflores-jahia:
fyi @GauBen from @TheKlem at today's daily, we should hold off merging content-versioning related changes for now on main until jcontent 3.7.1 has been released.

(If it starts to become a bottleneck, we can create a specific 3.7.1 branch, just lmk, but wouldn't want to do that extra work yet if we can hold off for a few days)

GauBen:
Fair enough, I added merging this PR at the end of the 3.7.1 release flow

